PR #2, opened 2025-04-01
Remove dependency

juecd:


containers/docker/Dockerfile
```diff
@@ -9,15 +9,6 @@ USER computeruse
 RUN curl -fsSL https://bun.sh/install | bash -s "bun-v1.1.44"
 ENV PATH="/home/computeruse/.bun/bin:$PATH"
```
rgarcia: can also remove bun install stuff I think

PR #3, opened 2025-04-06
unikernel config

rgarcia:
see https://github.com/onkernel/kernel/pull/50


unikernels/unikraft-cu/deploy.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
mertcelebi, line 1: This should be here and not in kernel-deployment right?

rgarcia: `kraft` makes this a hard call since there's no `kraft cloud build` command, only `kraft cloud deploy` which does both build + deploy. I think we keep it here since the deployment part is basically thrown away (actual deployments are done programmatically in the API via `kraft cloud instance create`) and all we care about is the image artifact it creates

PR #17, opened 2025-06-26
Implement replay server

Sayan-:
### Description

We want to be able to capture the screen for headful use cases. The path we chose to do this is simply:

1. Create an http server we'll embed into the image. This server will expose a set of endpoints to start, stop, and download recordings
2. We'll use this interface externally as necessary to capture the screen

For the moment we're encoding semantics that the server will have one "main" recording. I've implemented the rest of the backend without this restriction so it's easy to relax down the line. 

Under the hood this uses `ffmpeg` and expects that binary to be available. In tests running in docker and unikraft cloud the behavior around pause+resume was flaky. Instead of attempting to prevent data loss / corruption I took the strategy of minimizing the damage via ffmpeg fragmentation.

I setup the server using what felt like ~sane defaults. The meat of this PR is in `ffmpeg.go`. I recommend reviewing this commit by commit

### Testing

- [x] Added units tests where sensible
- [x] Setup github actions for aforementioned units
- [x] Performed a number of end to end tests both locally in docker + unikraft cloud using another locally running api to orchestrate the unikernels to replicate normal user workflows for persistent and non-persistent browsers


rgarcia (commented):
this is great. I have zero comments to add after reading through it!

I assume next step is figuring out how to incorporate a compiled binary of the server onto our docker / unikernel images?

PR #18, opened 2025-06-27
chore(config): move some neko cli args to yaml

xonkernel:
move some neko cli args to yaml


juecd:
@xonkernel I like the flag, but it's a little weird from a user experience perspective because it overrides the write toggle button (if you have ENABLE_READONLY_VIEW=true, the toggle toggles but doesn't actually change the setting, from my testing?).

Would it be possible to hide the toggle altogether? Then it's just controlled at start-up.

PR #19, opened 2025-06-27
Integrate api server

Sayan-:
### Description

In https://github.com/onkernel/kernel-images/pull/17 we built a new API server. In this PR we integrate it into the container + unikernel image builds. It's off by default and can be turned + configured with env vars

### Testing

- [x] Built docker container
- [x] Ran docker container locally without the server
- [x] Ran docker container locally with `WITH_KERNEL_IMAGES_API=true`. Confirmed the server started up and was able to perform the basic start, stop, down workflow successfully 
- [x] Built unikernel container
- [x] Ran unikernel in the cloud without the server
- [x] Ran unikernel in the cloud with `WITH_KERNEL_IMAGES_API=true`. Confirmed the server started up and was able to perform the basic start, stop, download workflow successfully


containers/docker/Dockerfile
```diff
@@ -2,11 +2,15 @@ FROM ghcr.io/anthropics/anthropic-quickstarts:computer-use-demo-latest
...
 RUN add-apt-repository -y ppa:xtradeb/apps
-RUN apt update -y && apt install -y chromium ncat
+RUN apt update -y && apt install -y chromium ncat ffmpeg
 
 # Switch to computeruse user
 USER computeruse
 
+# copy the kernel-images API binary built by build.sh
+COPY bin/kernel-images-api /usr/local/bin/kernel-images-api
+ENV WITH_KERNEL_IMAGES_API=false
```
rgarcia, line 12: should we build this in the container in a scratch image? e.g. https://github.com/onkernel/kernel/blob/main/packages/api/Dockerfile#L26-L31

containers/docker/Dockerfile
```diff
@@ -2,11 +2,15 @@ FROM ghcr.io/anthropics/anthropic-quickstarts:computer-use-demo-latest
```
rgarcia, line 2: discussed in standup but ideally we'd like to nuke the `containers/` directory so if you could merge the ideas in this `Dockerfile` with the one in the unikernels directory that would be great

Sayan-: Yeah it's copy pasted between the two! I'll touch base with cat on the nuking ^^

containers/docker/Dockerfile
Sayan-: I started down that route but since the `server` is external to the build context and I had some trouble getting multi-context builds working with `kraft pkg`. I could definitely be wrong and hope we can change it easily down the line!

PR #22, opened 2025-07-07
Fix docker build failure when buildkit container is absent

matthewjmarangoni:
# Checklist

- [ ] A link to a related issue in our repository
- [x] A description of the changes proposed in the pull request.
- [ ] @mentions of the person or team responsible for reviewing proposed changes.

## Description

Building the docker container fails when the _buildkit_ container does not already exist.

The failure is from a sourced script testing for container presence. The script exits when the test fails while having inherited the `-e` bash option.

The sourced presence test is modified. Additionally a typo is corrected, an EOF newline is added, and the sourcing script has bash tracing output disabled.

### DIFF
--- a/shared/start-buildkit.sh
+++ b/shared/start-buildkit.sh
@@ -6,7 +6,7 @@
 export KRAFTKIT_BUILDKIT_HOST=docker-container://buildkit
 
 # Install container if not already installed.
-if test "$(docker ps --all --no-trunc --quiet --filter 'name=^buildkit$')"; then
+if [ -n "$(docker ps --all --no-trunc --quiet --filter 'name=^buildkit$')" ]; then
     echo "Container 'buildkit' is already installed. Nothing to do."
 else
     echo "Installing 'buildkit' container ... "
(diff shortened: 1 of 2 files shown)


shared/start-buildkit.sh
```diff
@@ -6,8 +6,7 @@
 export KRAFTKIT_BUILDKIT_HOST=docker-container://buildkit
 
 # Install container if not already installed.
-docker container inspect buildkit > /dev/null 2>&1
-if test $? -eq 0; then
+if test "$(docker ps --all --no-trunc --quiet --filter 'name=^buildkit$')"; then
```
Sayan-:
```suggestion
if [ -n "$(docker ps --all --no-trunc --quiet --filter 'name=^buildkit$')" ]; then
```

shared/start-buildkit.sh
```diff
@@ -17,9 +16,9 @@ fi
 
 test "$(docker container inspect -f '{{.State.Running}}' buildkit 2> /dev/null)" = "true"
 if test $? -eq 0; then
```
Sayan-:
```suggestion
if [ "$(docker container inspect -f '{{.State.Running}}' buildkit 2> /dev/null)" = "true" ]; then
```

PR #29, opened 2025-07-11
Block scale to zero during recording

Sayan-:
As we've stress tested the recording server more we've noticed the behavior from pause <> resume to be a bit more unpredictable than we'd like. Instead we're opt'ing to leverage the pseudo file detailed in https://unikraft.cloud/docs/api/v1/instances/#scale-to-zero to disable scale in while we're recording. The interface doesn't change - the expectation is that the caller will still call `Stop`. We have size protection but in cases the caller doesn't explicitly call `Stop` the instance will stay `running` until it is torn town.

### Testing

Tested with the end to end automation suite. The expected behavior that the instance would stay in `running` until shutdown was explicitly called was confirmed


server/lib/scaletozero/scaletozero.go
```diff
@@ -0,0 +1,64 @@
...
+		logger.FromContext(ctx).Error("failed to open scale-to-zero control file", "path", c.path, "err", err)
+		return err
+	}
+	defer f.Close()
+	if _, err := f.Write([]byte(char)); err != nil {
+		logger.FromContext(ctx).Error("failed to write scale-to-zero control file", "path", c.path, "err", err)
+		return err
+	}
+	return nil
+}
```
Sayan-, line 58: see https://unikraft.cloud/docs/api/v1/instances/#scale-to-zero:~:text=Unikraft%20Cloud%20allows,file%2Dbased%20interface%3A

PR #34, opened 2025-07-14
clean up live view

rgarcia:
live view has much more chill now:

image

- no bookmark bar
- no fullscreen, resolution, or pip controls
- no "connected" toast
- no sound
- ability to set resolution and frame rate with query params e.g. `w=1024&h=1336&fr=60`
- ability to make the live view read-only (i.e. no mouse or keyboard input) with `readOnly=true`

Other nice things:
- set up dbus so no more log spam
- get rid of no-sandbox warning

recommend reviewing commit by commit or just ignoring everything in user-data/. In order to load custom preferences had to grab the default user-data directory that chromium creates and modify it / load it onto the image


images/chromium-headful/client/src/app.vue
```diff
@@ -251,6 +255,47 @@
...
+      const height = parseInt(params.get('height') || params.get('h') || '')
+      const rate = parseInt(params.get('rate') || params.get('r') || '30')
+
+      if (!isNaN(width) && !isNaN(height)) {
+        const resolution = { width, height, rate: isNaN(rate) ? 30 : rate }
+        this.$accessor.video.setResolution(resolution)
+        if (this.$accessor.user && this.$accessor.user.admin) {
+          this.$accessor.video.screenSet(resolution)
+        }
+      }
```
Sayan-: flyby question: what's `server` mean in this context? we aren't actually changing os level resolution are we? mainly interested in understanding the interaction between this and replays!

images/chromium-headful/client/src/app.vue
```diff
@@ -251,6 +255,47 @@
...
+    // and set the resolution accordingly. If the current user is an admin we also request the
+    // server to switch to that resolution.
+    private applyQueryResolution() {
+      const params = new URL(location.href).searchParams
+
+      const width = parseInt(params.get('width') || params.get('w') || '')
+      const height = parseInt(params.get('height') || params.get('h') || '')
+      const rate = parseInt(params.get('rate') || params.get('r') || '30')
+
+      if (!isNaN(width) && !isNaN(height)) {
```
heyglassy: nit: we should re-arrange this so that we check and handle if width, height and rate are NaN in one place.

heyglassy (approved):

images/chromium-headful/client/src/app.vue
rgarcia: "server" means the Neko backend process that the browser is talking to over a websocket (the same backend that streams the video and handles mouse/keyboard events). The backend uses libX11 to set the screen size: https://github.com/m1k1o/neko/blob/master/server/pkg/xorg/xorg.go#L200

images/chromium-headful/client/src/app.vue
rgarcia: a3be73a

images/chromium-headful/run-unikernel.sh
```diff
@@ -1,7 +1,41 @@
 #!/usr/bin/env bash
 
-image="onkernel/kernel-cu-test:latest"
-name="kernel-cu-test"
+source common.sh
+name=chromium-headful-test
+
+# Name for the Kraft Cloud volume that will carry Chromium flags
+volume_name="${name}-flags"
```
Sayan-: nit - the mix of upper and lower case vars is a little confusing

images/chromium-headful/run-unikernel.sh
```diff
@@ -1,7 +1,41 @@
...
+rm -rf .tmp/chromium
+mkdir -p .tmp/chromium
+FLAGS_DIR=".tmp/chromium"
+echo "$CHROMIUM_FLAGS" > "$FLAGS_DIR/flags"
+
+# Re-create the volume from scratch every run
+kraft cloud volume rm "$volume_name" || true
+kraft cloud volume create -n "$volume_name" -s 16M
+# Import the flags directory into the freshly created volume
+kraft cloud volume import -s "$FLAGS_DIR" -v "$volume_name"
```
Sayan-: do volume names need to be unique? mainly concerned about stepping on each others toes

images/chromium-headful/run-unikernel.sh
```diff
@@ -1,7 +1,41 @@
...
+
+# Ensure the temp directory is cleaned up on exit
+trap 'rm -rf "$FLAGS_DIR"' EXIT
+
+#   -e CHROMIUM_FLAGS="--user-data-dir=/home/kernel/user-data --disable-dev-shm-usage --disable-gpu --start-maximized --disable-software-rasterizer --remote-allow-origins=* --disable-breakpad --crash-dumps-dir=/tmp --no-sandbox --no-zygote"
```
Sayan-: should this be removed

server/cmd/api/api/computer.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/onkernel/kernel-images/server/lib/logger"
+	oapi "github.com/onkernel/kernel-images/server/lib/oapi"
+)
+
+func (s *ApiService) MoveMouse(ctx context.Context, request oapi.MoveMouseRequestObject) (oapi.MoveMouseResponseObject, error) {
```
Sayan-, line 12: non-blocking. there is a local dev flow for the server. since xdotool isn't supported on macos we may want to detect and return a 501 instead

images/chromium-headful/run-unikernel.sh
rgarcia: I think the convention that lives in my head is: lowercase for local vars, UPPERCASE for things related to env vars

not super consistent but can nudge things in that direction

images/chromium-headful/run-unikernel.sh
rgarcia: yeah... tricky part is ensuring we delete the previously-created volume (e.g. not sure how to `kraft cloud volume rm "$volume_name" || true` if we make volume name unique on every run)

images/chromium-headful/run-unikernel.sh
rgarcia: :nods:

server/cmd/api/api/computer.go
rgarcia: my local dev flow was building / running the docker image and curl'ing this endpoint... wasn't too bad

on mac os we might be able to do this with applescript

images/chromium-headful/run-unikernel.sh
Sayan-: yeah really good point - honestly this is where I usually just toss `$USER` in but that's not necessarily desirable 😅

Sayan- (approved):

images/chromium-headful/wrapper.sh
```diff
@@ -179,12 +202,13 @@ if [[ "${WITH_KERNEL_IMAGES_API:-}" == "true" ]]; then
...
-      echo "Waiting for kernel-images API port 10001..."
-      while ! nc -z localhost 10001 2>/dev/null; do
+      echo "Waiting for kernel-images API port 127.0.0.1:10001..."
+      while ! nc -z 127.0.0.1 10001 2>/dev/null; do
         sleep 0.5
       done
       echo "Port 10001 is open"
-
+      # wait... not sure but this just increases the likelihood of success
+      sleep 5
```
Sayan-: what were the failure modes you were running into?

images/chromium-headful/wrapper.sh
```diff
@@ -179,12 +202,13 @@ if [[ "${WITH_KERNEL_IMAGES_API:-}" == "true" ]]; then
       done
 
       # Wait for kernel-images API port 10001 to be ready.
-      echo "Waiting for kernel-images API port 10001..."
-      while ! nc -z localhost 10001 2>/dev/null; do
+      echo "Waiting for kernel-images API port 127.0.0.1:10001..."
+      while ! nc -z 127.0.0.1 10001 2>/dev/null; do
```
Sayan-: potential optimization: I'd expect the server to come up faster than chromium so we could probably trim some time by inverting the order

images/chromium-headful/build-unikernel.sh
```diff
@@ -1,20 +1,15 @@
...
-    exit 1
-fi
 source ../../shared/start-buildkit.sh
 
 # Build the API binary
 source ../../shared/build-server.sh "$(pwd)/bin"
 
 kraft pkg \
-  --name index.unikraft.io/$image \
+  --name $UKC_INDEX/$IMAGE \
```
Sayan-, line 12: should we default to the public unikraft index to make this easier for folks?

Sayan- (approved):

images/chromium-headful/build-unikernel.sh
rgarcia: 👍

images/chromium-headful/wrapper.sh
rgarcia: 👍

images/chromium-headful/wrapper.sh
rgarcia: without the sleep I'd often open the live view and see the mouse hovering over the "X" to dismiss the warning, suggesting that xdotool did it's thing before the warning or chromium appeared

will add to the comment describing this

Sayan- (approved):

Issue #28, opened 2025-07-10
WebRTC shows double cursors


juecd:
WebRTC shows a second cursor on the GUI, which largely follows your cursor but has a bit of latency. It's a bit jarring to see multiple cursors.

raiden-staging:
tfw [fix< issue length](https://github.com/raiden-staging/kernel-images/blob/1cbe8e0e89f2a08d37281dbbf41cebeb1b4a2281/images/chromium-headful/client/public/index.html#L16) 😄

PR #35, opened 2025-07-17
remove live view duplicate cursor

raiden-staging:
quick issue #28 fix

image


images/chromium-headful/client/bun.lockb
rgarcia, line 1: remove this file please!

juecd (approved):

Issue #31, opened 2025-07-12
Copy/paste doesn't work


juecd:
Copy paste doesn't work with WebRTC!

raiden-staging:
clipboard events+payload are logging in console. looks like sockets well sent+received but not working on neko's end.
- in worst case , can have the vue client write to clipboard and call event to handle outside of neko on another open port
- (added a temporary workaround on vue that sends key sequence to help remote debug in case)

juecd:
Note here for possible further investigation: https://discord.com/channels/1342243238748225556/1394352259847950378/1394353221652316327

raiden-staging:
latest merge [ #46 ] uses custom neko base where clipboard was previously fixed
so clipboard works now (on `api.onkernel.com/browser/live` too) ✅

raiden-staging:
https://github.com/user-attachments/assets/76b34c7b-7510-4faa-a097-91c8775241cc

- works on chrome
- doesn't work on firefox, see [`neko/client/src/components/video.vue`](https://github.com/m1k1o/neko/blob/96d9e8a6ea789d855cb134308237e2c5dcae7514/client/src/components/video.vue#L332)

notes : 
- should add firefox workaround that triggers on paste only
- should also add virtual keyboard to vue client to control session from mobile later

PR #36, opened 2025-07-18
copy/paste fix (v1)

raiden-staging:
(Issue #31)

https://github.com/user-attachments/assets/2d0b7d60-850b-47c8-9355-dd904c86db80

---

## Issue Trace

- Running container logs on clipboard events

```bash
9:29PM WRN message handler has failed error="is not the host" event=clipboard/set module=websocket session_id=dummy-vPhbw submodule=handler
```

- `onkernel/kernel-images:images/chromium-headful/Dockerfile` :

```Dockerfile
FROM ghcr.io/m1k1o/neko/chromium:3.0.6 AS neko
```

- `m1k1o/neko:server/internal/websocket/handler/clipboard.go` | [see here](https://github.com/m1k1o/neko/blob/1bc31473aeeef145937ad7723c0beb9b1a5ec80d/server/internal/websocket/handler/clipboard.go)

```
func (h *MessageHandlerCtx) clipboardSet(session types.Session, payload *message.ClipboardData) error {
	if !session.Profile().CanAccessClipboard {
		return errors.New("cannot access clipboard")
	}

	if !session.IsHost() {
		return errors.New("is not the host")
	}
...
```

- example member profile config | [see here](https://neko.m1k1o.net/docs/v3/configuration/authentication#session)

```yaml
name: User Name
is_admin: false
...
can_access_clipboard: true # <-----------------------------
...
```

---

### Alternative A (not tried)

- in `onkernel/kernel-images:images/chromium-headful/client/src/components/connect.vue`
```
...
	// KERNEL: auto-login
	this.$accessor.login({ displayname: 'dummy', password: 'dummy' })
	this.autoPassword = null
```

- suggested change would be setting up neko profiles and passing them during build steps
- note : needs to consider multiplayer user scenarios (user/user and user/agent) for later , and control scopes based on roles , which i assume would be assigned outside of the chromium container and passed to it from the kernel APIs
 
### Alternative B (implemented)

- added `POST /computer/paste : { text }` to the Go `/server` to manage clipboard + trigger paste , enabled CORS to be called by the web client
- added a global plugin on the `vue` client to capture clipboard paste events and call the local server

note
- separate APIs for managing IO seems a good approach for better control by computer use / agent SDKs later [?]
- important : use WITH_KERNEL_IMAGES_API
  - `WITH_KERNEL_IMAGES_API=true ENABLE_WEBRTC=true ./run-docker.sh`
---

<span data-mesa-description="start"></span>
## TL;DR
Fixed clipboard paste functionality for non-host clients by introducing a new local API endpoint and client-side paste interception.

## Why we made these changes
The existing `neko` clipboard mechanism restricted paste operations to the "host" session, preventing other clients (like the dummy client in the chromium container) from pasting content, as evidenced by the "is not the host" error. This change enables universal paste functionality.

## What changed?
- **Server:**
    - Added `POST /computer/paste` endpoint to handle clipboard setting and simulate `Ctrl+V` (using `xclip` and `xdotool`).
    - Enabled CORS for the API server to allow client-side calls.
    - Added `go-chi/cors` dependency and updated OpenAPI spec.
- **Client:**
    - Introduced `GlobalPaste` Vue plugin to intercept global paste events (Ctrl+V/Cmd+V).
    - Sends intercepted clipboard text to the new local `/computer/paste` endpoint.
    - Updated `video.vue` and `remote.ts` to use the new local fallback service for clipboard updates.
    - Added minor logging and updated VS Code settings.

## Validation
Verified with provided video demonstration (Issue #31).
<span data-mesa-description="end"></span>


images/chromium-headful/client/src/plugins/globalPaste.ts
```diff
@@ -0,0 +1,30 @@
...
+    document.addEventListener(
+      'keydown',
+      async (e: KeyboardEvent) => {
+        if ((e.ctrlKey || e.metaKey) && e.key.toLowerCase() === 'v') {
+          e.preventDefault()
+          console.log(`[vue:globalPaste]:call`)
+          try {
+            const text = await navigator.clipboard.readText()
+            console.log(`[vue:globalPaste]:payload:` , text)
+            await fetch('http://localhost:10001/computer/paste', {
```
rgarcia, line 15: using localhost here won't work in prod

rgarcia (changes requested):
I think "Alternative A" is more straightforward, i.e. giving the "dummy" user a profile that allows clipboard use in neko

images/chromium-headful/client/src/plugins/globalPaste.ts
raiden-staging: on it 👍

note :
idea of external controls (outside of neko events) exposed via API might be useful down the line ; as a controller for agents to stream events to/from via sockets. not via localhost on the instance like in here - would be routed to and token-gated by the kernel api , and connected to by the ^ container and user/agent outside of the vue app

rgarcia: yes definitely--this is why we added the click/move mouse stuff into the API. I think eventually we will go deeper on that front

PR #39, opened 2025-07-22
use erofs for headful image

rgarcia:
syncing up with the approach used in the headless image

this should reduce memory usage

---

<span data-mesa-description="start"></span>
## TL;DR
Switched the Chromium headful unikernel to use EROFS for its root filesystem, aligning it with the headless image and reducing memory usage.

## Why we made these changes
To reduce memory consumption for the Chromium headful image and standardize its root filesystem approach with the existing headless image, improving overall efficiency.

## What changed?
*   **Chromium Headful Image:**
    *   Updated `Kraftfile` to use `idle` scale-to-zero policy, increased cooldown, and pointed rootfs to `./initrd`.
    *   Modified `build-unikernel.sh` to generate the `initrd` using `mkfs.erofs` from Docker-extracted contents.
*   **Shared Utilities:**
    *   Introduced `shared/erofs-utils.sh` with functions to check for and install `mkfs.erofs`.
    *   Refactored `chromium-headless/build-unikernel.sh` to utilize these new shared utility scripts, removing duplicated logic.

## Validation
*   [ ] Chromium headful unikernel builds successfully.
*   [ ] Headful image boots and functions as expected.
*   [ ] Memory usage is reduced compared to the previous build.
*   [ ] Scale-to-zero and cooldown policies in `Kraftfile` behave correctly.
<span data-mesa-description="end"></span>

### DIFF
--- a/images/chromium-headful/.gitignore
+++ b/images/chromium-headful/.gitignore
@@ -1,3 +1,5 @@
 bin/
 recording/
 .tmp/
+.rootfs/
+initrd
--- a/images/chromium-headful/Kraftfile
+++ b/images/chromium-headful/Kraftfile
@@ -3,10 +3,10 @@ spec: v0.6
 runtime: index.unikraft.io/official/base-compat:latest
 
 labels:
-  cloud.unikraft.v1.instances/scale_to_zero.policy: "on"
+  cloud.unikraft.v1.instances/scale_to_zero.policy: "idle"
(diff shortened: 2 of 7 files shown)


Sayan- (approved):
code changes lgtm. I'd like to understand how we want to verify the mem footprint savings but maybe it's a ship it and see situation

PR #44, opened 2025-08-03
file i/o endpoints

rgarcia:
- read/write individual files
- list / create directories
- move files
- set file permissions
- watch for changes

idea is to expose most if not all of these in the main API


server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,488 @@
...
+		if err != nil {
+			log.Error("failed to stat directory entry", "err", err, "dir", path, "entry", entry.Name())
+			return oapi.ListFiles500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "failed to stat directory entry"}}, nil
+		}
+
+		// By specification SizeBytes should be 0 for directories.
+		size := 0
+		if !info.IsDir() {
+			size = int(info.Size())
+		}
```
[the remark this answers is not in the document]
rgarcia: the 2000s are over, we will never again compile this to target a 32 bit architecture

server/openapi.yaml
```diff
@@ -154,6 +154,297 @@ paths:
...
+              $ref: "#/components/schemas/SetFilePermissionsRequest"
+      responses:
+        "200":
+          description: Permissions updated
+        "400":
+          $ref: "#/components/responses/BadRequestError"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+        "500":
+          $ref: "#/components/responses/InternalError"
```
Sayan-: I think all these `POST`'s could reasonably be `PUT`s - mainly calling it out for retries since ~most clients won't retry a POST vs a PUT

images/chromium-headful/run-docker.sh
```diff
@@ -45,7 +43,7 @@ RUN_ARGS=(
 )
 
 if [[ "${WITH_KERNEL_IMAGES_API:-}" == "true" ]]; then
-  RUN_ARGS+=( -p 10001:10001 )
+  RUN_ARGS+=( -p 444:10001 )
```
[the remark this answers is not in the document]
rgarcia: run-docker.sh works for me w/o root

server/openapi.yaml
rgarcia: good call 549e537

server/openapi.yaml
```diff
@@ -154,6 +154,297 @@ paths:
           $ref: "#/components/responses/BadRequestError"
         "500":
           $ref: "#/components/responses/InternalError"
+  # File system operations
```
Sayan-, line 157: 🔥

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+		log.Error("failed to open file", "err", err, "path", path)
+		return oapi.ReadFile400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "unable to open file"}}, nil
+	}
+
+	stat, err := f.Stat()
+	if err != nil {
+		f.Close()
+		log.Error("failed to stat file", "err", err, "path", path)
+		return oapi.ReadFile400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "unable to stat file"}}, nil
```
Sayan-: We should 500 here

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+		log.Error("failed to create directories", "err", err, "path", path)
+		return oapi.WriteFile400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "unable to create directories"}}, nil
+	}
+
+	// determine desired file mode (default 0o644)
+	perm := os.FileMode(0o644)
+	if req.Params.Mode != nil {
+		if v, err := strconv.ParseUint(*req.Params.Mode, 8, 32); err == nil {
+			perm = os.FileMode(v)
+		}
```
Sayan-: Should we 400 here? Silently failing ~shouldn't really cause issues in our environment but I could go either way for correctness

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+	if err != nil {
+		log.Error("failed to create file", "err", err, "path", path)
+		return oapi.WriteFile400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "unable to create file"}}, nil
+	}
+	defer f.Close()
+
+	if _, err := io.Copy(f, req.Body); err != nil {
+		log.Error("failed to write file", "err", err, "path", path)
+		return oapi.WriteFile400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "failed to write data"}}, nil
+	}
```
Sayan-: Should 500 here as well

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+	if path == "" {
+		return oapi.CreateDirectory400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "path cannot be empty"}}, nil
+	}
+	// default to 0o755
+	perm := os.FileMode(0o755)
+	if req.Body.Mode != nil {
+		if v, err := strconv.ParseUint(*req.Body.Mode, 8, 32); err == nil {
+			perm = os.FileMode(v)
+		}
+	}
```
Sayan-: ditto on mode

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+		// Retrieve FileInfo for each entry. If this fails (e.g. broken symlink, permission
+		// error) we surface the failure to the client instead of silently ignoring it so
+		// that consumers do not unknowingly operate on incomplete or unreliable metadata.
+		info, err := entry.Info()
+		if err != nil {
+			log.Error("failed to stat directory entry", "err", err, "dir", path, "entry", entry.Name())
+			return oapi.ListFiles500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "failed to stat directory entry"}}, nil
+		}
+
+		// By specification SizeBytes should be 0 for directories.
```
Sayan-, line 219: 👍

server/cmd/api/api/fs.go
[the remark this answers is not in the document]
Sayan-: based

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	fi := oapi.FileInfo{
+		Name:      filepath.Base(path),
+		Path:      path,
+		IsDir:     stat.IsDir(),
+		SizeBytes: int(stat.Size()),
+		ModTime:   stat.ModTime(),
+		Mode:      stat.Mode().String(),
+	}
+	return oapi.FileInfo200JSONResponse(fi), nil
```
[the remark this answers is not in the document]
Sayan-: == to this difference

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+	path := req.Body.Path
+	if path == "" {
+		return oapi.StartFsWatch400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "path cannot be empty"}}, nil
+	}
+	// Ensure path exists
+	if _, err := os.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return oapi.StartFsWatch404JSONResponse{NotFoundErrorJSONResponse: oapi.NotFoundErrorJSONResponse{Message: "path not found"}}, nil
+		}
+		return oapi.StartFsWatch400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "unable to stat path"}}, nil
```
Sayan-: Should 500 here

server/cmd/api/api/fs.go
```diff
@@ -0,0 +1,521 @@
...
+				return
+			}
+			if err := enc.Encode(ev); err != nil {
+				log.Error("failed to encode fs event", "err", err)
+				return
+			}
+			if _, err := pw.Write([]byte("\n")); err != nil {
+				log.Error("failed to write SSE terminator", "err", err)
+				return
+			}
```
Sayan-: I wonder if this would be more efficient as some string interpolation with 1 `pw.Write` instead of 3 writes. I don't think this necessarily becomes 3 network calls but it's a little strange and I'm not sure how to describe it 😅

Sayan- (changes requested):
Overall API + flows lgtm! Some errors I think we should fix but nothing super deep.

The main question I have is whether we should configure basic allow list / deny list shaped configuration so folks don't read/mv/delete things that could cause serious issues. I don't think we can prevent event failure mode but I think we could broadly guard against the cases we care about. What do you think?

shared/cdp-test/main.py
```diff
@@ -3,18 +3,109 @@
...
+                                content = await file_resp.read()
+                                local_file_path = local_download_path / file_name
+                                with open(local_file_path, "wb") as f:
+                                    f.write(content)
+                                print(f"Downloaded file saved to {local_file_path.resolve()}")
+                            else:
+                                error_text = await file_resp.text()
+                                print(f"Failed to fetch {file_name}: {file_resp.status} - {error_text}", file=sys.stderr)
+
+async def run(cdp_url: str | None = None) -> None:
```
Sayan-, line 100: this is quite nifty

Sayan- (approved):
Confirming my allow list / deny list isn't blocking for this pr! Let's go 🚀 !

PR #46, opened 2025-08-04
peer disconnections

rgarcia:
#45 plus 
- switching to use our build of the neko image
- bringing back double cursor (lowers perceived latency)

thanks very much to @raiden-staging !

### DIFF
--- a/images/chromium-headful/client/src/neko/index.ts
+++ b/images/chromium-headful/client/src/neko/index.ts
@@ -149,7 +149,12 @@ export class NekoClient extends BaseClient implements EventEmitter<NekoEvents> {
 
     if (heartbeat_interval > 0) {
       if (this._ws_heartbeat) clearInterval(this._ws_heartbeat)
-      this._ws_heartbeat = window.setInterval(() => this.sendMessage(EVENT.CLIENT.HEARTBEAT), heartbeat_interval * 1000)
+      this._ws_heartbeat = window.setInterval(() => {
+        this.emit('debug', `sending client/heartbeat`)
+        this.sendMessage(EVENT.CLIENT.HEARTBEAT)
+        this.emit('debug', `sending chat/message`)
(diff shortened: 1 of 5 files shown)


images/chromium-headful/client/src/neko/index.ts
```diff
@@ -149,7 +149,12 @@ export class NekoClient extends BaseClient implements EventEmitter<NekoEvents> {
 
     if (heartbeat_interval > 0) {
       if (this._ws_heartbeat) clearInterval(this._ws_heartbeat)
-      this._ws_heartbeat = window.setInterval(() => this.sendMessage(EVENT.CLIENT.HEARTBEAT), heartbeat_interval * 1000)
+      this._ws_heartbeat = window.setInterval(() => {
+        this.emit('debug', `sending client/heartbeat`)
+        this.sendMessage(EVENT.CLIENT.HEARTBEAT)
+        this.emit('debug', `sending chat/message`)
+        this.sendMessage(EVENT.CHAT.MESSAGE, { content: `heartbeat/fake [${Date.now()}]` })
```
Sayan-, line 156: @raiden-staging  - just for context what's the motivation of this additional message type?

I thought we would be sufficient with just the heartbeat + server side response you added!

Sayan- (approved):

images/chromium-headful/client/src/neko/index.ts
raiden-staging: ( @Sayan-  @rgarcia )
- `event.CLIENT_HEARTBEAT` triggers new `event.SYSTEM_HEARTBEAT` reply from neko, but doesn't show up on websockets trace. will fix on neko/ws handler + new neko build in few.
- `event.CHAT_MESSAGE → reply` is handling keepalive for now.
- unikernel deployment seems to work well so cleanup for the above incoming 🧹✨

Sayan-: Incredible - thank you again for digging into this issue!

PR #54, opened 2025-08-16
[Extending API] /screenshot

raiden-staging:
- Extending `server` with `/screenshot` endpoint
```bash
# Screenshot

# | GET /screenshot
curl http://localhost:10001/screenshot -o screenshot.png

# | POST /screenshot
curl -X POST -H "Content-Type: application/json" --data '{"x":0,"y":0,"width":200,"height":200}' -o screenshot_region.png http://localhost:10001/screenshot
```


server/cmd/api/api/screenshot.go
```diff
@@ -0,0 +1,195 @@
...
+					if w > 0 && h > 0 {
+						return w, h, nil
+					}
+				}
+			}
+		}
+	}
+
+	// Default dimensions if parsing fails
+	return 1024, 768, nil
```
matthewjmarangoni, line 47: If the fallback is needed would it be useful to read fallback dimensions from something such as environment variables before fixed values?

server/lib/oapi/oapi.go
```diff
@@ -1,6 +1,6 @@
 // Package oapi provides primitives to interact with the openapi HTTP API.
 //
-// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.5.0 DO NOT EDIT.
+// Code generated by github.com/deepmap/oapi-codegen version v1.12.4 DO NOT EDIT.
```
rgarcia, line 3: this doesn't look right

server/openapi.yaml
```diff
@@ -3,6 +3,54 @@ info:
...
+          description: Screenshot captured successfully
+          content:
+            image/png:
+              schema:
+                type: string
+                format: binary
+        "400":
+          $ref: "#/components/responses/BadRequestError"
+        "500":
+          $ref: "#/components/responses/InternalError"
```
rgarcia, line 53: I think I'd prefer merging these into one POST request

server/cmd/api/api/screenshot.go
```diff
@@ -0,0 +1,195 @@
...
+		log.Error("failed to capture screenshot", "error", err)
+		return oapi.CaptureScreenshot500JSONResponse{
+			InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{
+				Message: err.Error(),
+			},
+		}, nil
+	}
+
+	// Read the screenshot data
+	data, err := io.ReadAll(screenshot)
```
rgarcia, line 131: not sure how oapi does response body reading under the hood but I'd assume it's doing this read of the reader anyway, so this feels like duplicate / unnecessary reading of the screenshot into memory

server/lib/oapi/oapi.go
raiden-staging: oapi/go version conflicts fix induced slop 🙃 sorry for the mess will fix

server/cmd/api/api/screenshot.go
raiden-staging: should we serve as base64 string [ @matthewjmarangoni ] or binary (or add optional `?format={}` ) ?

matthewjmarangoni: If I understand the question correctly, offering both with a way for the client to choose either the original binary or base 64 encoded string is what I had in mind.

PR #61, opened 2025-08-23
[Extending API] /screen/resolution

raiden-staging:
[ @rgarcia ]

---
- Added `/screen/resolution` to server, to update the neko screen resolution from the API

```bash
# | GET /screen/resolution
# | required : width,height
# | optional : rate
curl http://localhost:10001/screen/resolution?width=720&height=1280&rate=60
```

- checks for ongoing ffmpeg processes to avoid interruptions in case replay is being recorded
- uses neko websockets , adresses [`onkernel/neko:#3`](https://github.com/onkernel/neko/issues/3) without changes to neko

### DIFF
--- a/server/cmd/api/api/api.go
+++ b/server/cmd/api/api/api.go
@@ -83,31 +83,42 @@ func isWebSocketAvailable(wsURL string) bool {
 // GetWebSocketURL determines the appropriate WebSocket URL from an HTTP request
 // It can be used in tests
 func getWebSocketURL(r *http.Request) string {
-	// Default fallback URL for local development with auth parameters
-	localURL := "ws://localhost:8080/ws?password=admin&username=kernel"
+	// Auth parameters for WS connection
+	authParams := "?password=admin&username=kernel"
+
+	// Default local development URL - will try only in local dev
+	localDevURL := "ws://localhost:8080/ws" + authParams
 
 	// In tests or other cases where request is nil
 	if r == nil {
-		return localURL
+		return localDevURL
 	}
 
-	// Try the local URL first
-	if isWebSocketAvailable(localURL) {
-		log := logger.FromContext(r.Context())
-		log.Info("using local WebSocket URL", "url", localURL)
-		return localURL
-	}
+	log := logger.FromContext(r.Context())
 
-	// Local URL not available, build URL from request
+	// Get URL components from the request
 	scheme := "ws"
-	if r.TLS != nil {
+	if r.TLS != nil || strings.HasPrefix(r.Proto, "HTTPS") || r.Header.Get("X-Forwarded-Proto") == "https" {
 		scheme = "wss"
 	}
 
-	// Get host from request header
+	// Get host from request header, strip the port if present
+	// This is crucial for production where we don't want ports in WS URLs
 	host := r.Host
 	if host == "" {
-		return localURL // Fall back to local if host is missing
+		log.Warn("empty host in request, using fallback mechanisms")
+
+		// Try the internal WebSocket endpoint
+		internalURL := "ws://127.0.0.1:8080/ws" + authParams
+		log.Info("trying internal WebSocket URL", "url", internalURL)
+
+		// If it fails, return the URL anyway since we need to return something
+		return internalURL
+	}
+
+	// Remove port from host if present (critical for production)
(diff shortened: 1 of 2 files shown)


server/openapi.yaml
```diff
@@ -737,6 +737,53 @@ paths:
           $ref: "#/components/responses/NotFoundError"
         "500":
           $ref: "#/components/responses/InternalError"
+          
+  /screen/resolution:
+    get:
```
matthewjmarangoni, line 742: POST may be more appropriate for modifying the screen state.

Adding a GET to retrieve the current screen state might be useful too.

server/cmd/api/api/api.go
```diff
@@ -28,8 +35,200 @@ type ApiService struct {
 	procs  map[string]*processHandle
 }
 
+// We're extending the StrictServerInterface to include our new endpoint
 var _ oapi.StrictServerInterface = (*ApiService)(nil)
 
+// SetScreenResolution endpoint
+// (GET /screen/resolution)
```
matthewjmarangoni, line 42: Possible temporary note or outdated comment.

server/cmd/api/api/api.go
```diff
@@ -28,8 +31,90 @@ type ApiService struct {
...
+		w.WriteHeader(http.StatusConflict)
+		json.NewEncoder(w).Encode(r)
+	case SetScreenResolution500JSONResponse:
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(r)
+	default:
+		http.Error(w, "unexpected response type", http.StatusInternalServerError)
+	}
+}
```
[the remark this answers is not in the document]
rgarcia: @raiden-staging cursor is correct here. The rough process to add a new endpoint:
- add it to openapi.yaml
- run `make oapi-generate`
- implement the new method generated on StrictServerInterface

server/cmd/api/api/computer.go
```diff
@@ -54,6 +60,226 @@ func (s *ApiService) MoveMouse(ctx context.Context, request oapi.MoveMouseReques
...
+
+	// Prepare multiple fallback URLs for different environments
+	fallbackURLs := []string{
+		// Internal container URL (direct to the WS service)
+		"ws://127.0.0.1:8080/ws?password=admin&username=kernel",
+		// Docker service name URL (for container networking)
+		"ws://browser:8080/ws?password=admin&username=kernel",
+		// Local development fallback
+		"ws://localhost:8080/ws?password=admin&username=kernel",
+	}
```
rgarcia, line 159: 127.0.0.1 should always work, right? We run this server alongside neko on the same instance

server/cmd/api/api/computer.go
```diff
@@ -54,6 +60,226 @@ func (s *ApiService) MoveMouse(ctx context.Context, request oapi.MoveMouseReques
...
+
+	// Check if ffmpeg is running (indicating an active recording)
+	cmd := exec.Command("pgrep", "ffmpeg")
+	if err := cmd.Run(); err == nil {
+		// ffmpeg is running
+		return SetScreenResolution409JSONResponse{
+			Message: "detected ongoing replay recording process, close the recording first before switching resolution",
+		}, nil
+	}
```
rgarcia, line 146: cc @Sayan- on ideas for how this should interact with replays

server/openapi.yaml
rgarcia: agreed!

server/cmd/api/api/computer.go
raiden-staging: - should handle relaying to neko ws in different scenarios, resolves from internal / exposed / host subdomain
- i tested on deployed unikraft with server+neko on same instance; but please confirm the api case

server/cmd/api/api/computer.go
Sayan-: I think instead of a `pgrep` you could fetch the list of recorders and determine if we're actively recording via:

```golang
	recs := s.recordManager.ListActiveRecorders(ctx)
	for _, r := range recs {
		if r.IsRecording(ctx) {
			return true
		}
	}
	return false
```

I'd also recommend grabbing a lock and blocking `StartRecording` until we've completed the screen resizing

PR #62, opened 2025-08-25
upd(novnc): remove old references

matthewjmarangoni:
Clean up some references to noVNC since it's no longer installed. Please close if this is irrelevant.


rgarcia:
thanks @matthewjmarangoni !

PR #66, opened 2025-09-10
websocket disconnections

rgarcia:
- changed run-unikernel scripts to match resources used in prod
- added an env var flag LOG_CDP_MESSAGES that will log a condensed form of each CDP message
- removed the heartbeat ping. Turns out browser use sometimes asks a /lot/ of the CDP connection when it first begins. It pulls the accessibilty tree plus full document which can be many MBs. Was seeing this ping time out during this initial startup, causing us to close the connection. 
- adjusted websocket dialer and connection to be very lenient
- changed the uk-check-stats script to compute a real cpu % and a KB/s network rate

todo:
- [ ] test with API in the middle
- [ ] test keeping the ping but not killing the connection if it times out


rgarcia:
1. From my reading it seems like the behavior when those buffers get filled is network lag--basically the buffer gets filled and then the connection stops reading data. Making them big hopefully accommodates big spikes in network activity with less lag
2. I think the overhead is pretty minimal of compression but yes it depends on the client asking for it (in our case, the API). I mainly enabled it to minimize stress on the network. I was not seeing CPU limit (4) getting hit in my observations of uk-check-stats output

PR #73, opened 2025-09-30
WIP: POC for configurable viewport 

hiroTamada:
Initially
image

After running
```
➜  kernel-images git:(hiro/configurable_viewport) curl -X POST http://localhost:444/display/set_resolution \
  -H "Content-Type: application/json" \
  -d '{"width": 1920, "height": 1080, "require_idle": false}'
{"ok":true}
```
image

### DIFF
--- a/images/chromium-headful/Dockerfile
+++ b/images/chromium-headful/Dockerfile
@@ -124,6 +124,7 @@ RUN set -eux; \
     wget ca-certificates python2 supervisor xclip xdotool \
     pulseaudio dbus-x11 xserver-xorg-video-dummy \
     libcairo2 libxcb1 libxrandr2 libxv1 libopus0 libvpx7 \
+    x11-xserver-utils \
     gstreamer1.0-plugins-base gstreamer1.0-plugins-good \
     gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly \
     gstreamer1.0-pulseaudio gstreamer1.0-omx; \
--- a/server/cmd/api/api/display.go
+++ b/server/cmd/api/api/display.go
@@ -1,0 +1,162 @@
+package api
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/onkernel/kernel-images/server/lib/logger"
+	oapi "github.com/onkernel/kernel-images/server/lib/oapi"
+)
+
+// DisplayStatus reports whether it is currently safe to resize the display.
+// It checks for active Neko viewer sessions (approx. by counting ESTABLISHED TCP
+// connections on port 8080) and whether any recording is active.
+func (s *ApiService) DisplayStatus(ctx context.Context, _ oapi.DisplayStatusRequestObject) (oapi.DisplayStatusResponseObject, error) {
+	live := s.countEstablishedTCPSessions(ctx, 8080)
+	isRecording := s.anyRecordingActive(ctx)
+	isReplaying := false // replay not currently implemented
+
+	resizableNow := (live == 0) && !isRecording && !isReplaying
+
+	return oapi.DisplayStatus200JSONResponse(oapi.DisplayStatus{
+		LiveViewSessions: &live,
+		IsRecording:      &isRecording,
+		IsReplaying:      &isReplaying,
+		ResizableNow:     &resizableNow,
+	}), nil
+}
+
+// SetResolution safely updates the current X display resolution. When require_idle
+// is true (default), it refuses to resize while live view or recording/replay is active.
+func (s *ApiService) SetResolution(ctx context.Context, req oapi.SetResolutionRequestObject) (oapi.SetResolutionResponseObject, error) {
+	log := logger.FromContext(ctx)
+	if req.Body == nil {
+		return oapi.SetResolution400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "missing request body"}}, nil
+	}
+	width := req.Body.Width
+	height := req.Body.Height
+	if width <= 0 || height <= 0 {
+		return oapi.SetResolution400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "invalid width/height"}}, nil
+	}
+	requireIdle := true
+	if req.Body.RequireIdle != nil {
+		requireIdle = *req.Body.RequireIdle
+	}
+
+	// Check current status
+	statusResp, _ := s.DisplayStatus(ctx, oapi.DisplayStatusRequestObject{})
+	var status oapi.DisplayStatus
+	switch v := statusResp.(type) {
+	case oapi.DisplayStatus200JSONResponse:
+		status = oapi.DisplayStatus(v)
+	default:
+		// In unexpected cases, default to conservative behaviour
+		status = oapi.DisplayStatus{LiveViewSessions: ptrInt(0), IsRecording: ptrBool(false), IsReplaying: ptrBool(false), ResizableNow: ptrBool(true)}
+	}
+	if requireIdle && status.ResizableNow != nil && !*status.ResizableNow {
+		return oapi.SetResolution409JSONResponse{ConflictErrorJSONResponse: oapi.ConflictErrorJSONResponse{Message: "resize refused: live view or recording/replay active"}}, nil
+	}
+
+	display := s.resolveDisplayFromEnv()
+	args := []string{"-lc", fmt.Sprintf("xrandr -s %dx%d", width, height)}
+	env := map[string]string{"DISPLAY": display}
+	execReq := oapi.ProcessExecRequest{Command: "bash", Args: &args, Env: &env}
+	resp, err := s.ProcessExec(ctx, oapi.ProcessExecRequestObject{Body: &execReq})
+	if err != nil {
+		return oapi.SetResolution500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "failed to execute xrandr"}}, nil
+	}
+	switch r := resp.(type) {
+	case oapi.ProcessExec200JSONResponse:
+		if r.ExitCode != nil && *r.ExitCode != 0 {
+			var stderr string
+			if r.StderrB64 != nil {
+				if b, decErr := base64.StdEncoding.DecodeString(*r.StderrB64); decErr == nil {
+					stderr = strings.TrimSpace(string(b))
+				}
+			}
+			if stderr == "" {
+				stderr = "xrandr returned non-zero exit code"
+			}
+			return oapi.SetResolution400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: fmt.Sprintf("failed to set resolution: %s", stderr)}}, nil
+		}
+		log.Info("resolution updated", "display", display, "width", width, "height", height)
+
+		// Restart Chromium to ensure it adapts to the new resolution
+		log.Info("restarting chromium to adapt to new resolution")
+		restartCmd := []string{"-lc", "supervisorctl restart chromium"}
+		restartEnv := map[string]string{}
+		restartReq := oapi.ProcessExecRequest{Command: "bash", Args: &restartCmd, Env: &restartEnv}
+		restartResp, restartErr := s.ProcessExec(ctx, oapi.ProcessExecRequestObject{Body: &restartReq})
+		if restartErr != nil {
+			log.Error("failed to restart chromium after resolution change", "error", restartErr)
+			// Still return success since resolution change succeeded
+			return oapi.SetResolution200JSONResponse{Ok: true}, nil
+		}
+
+		// Check if restart succeeded
+		if execResp, ok := restartResp.(oapi.ProcessExec200JSONResponse); ok {
+			if execResp.ExitCode != nil && *execResp.ExitCode != 0 {
+				log.Error("chromium restart failed", "exit_code", *execResp.ExitCode)
+			} else {
+				log.Info("chromium restarted successfully")
+			}
+		}
+
+		return oapi.SetResolution200JSONResponse{Ok: true}, nil
+	case oapi.ProcessExec400JSONResponse:
+		return oapi.SetResolution400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: r.Message}}, nil
+	case oapi.ProcessExec500JSONResponse:
+		return oapi.SetResolution500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: r.Message}}, nil
+	default:
+		return oapi.SetResolution500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "unexpected response from process exec"}}, nil
+	}
+}
+
+// anyRecordingActive returns true if any registered recorder is currently recording.
+func (s *ApiService) anyRecordingActive(ctx context.Context) bool {
+	for _, r := range s.recordManager.ListActiveRecorders(ctx) {
+		if r.IsRecording(ctx) {
+			return true
+		}
+	}
+	return false
+}
+
+// countEstablishedTCPSessions returns the number of ESTABLISHED TCP connections for the given local port.
+// Implementation shells out to netstat, which is present in the image (net-tools).
+func (s *ApiService) countEstablishedTCPSessions(ctx context.Context, port int) int {
+	cmd := exec.CommandContext(ctx, "/bin/bash", "-lc", fmt.Sprintf("netstat -tn 2>/dev/null | awk '$6==\"ESTABLISHED\" && $4 ~ /:%d$/ {count++} END{print count+0}'", port))
+	out, err := cmd.Output()
+	if err != nil {
+		return 0
+	}
+	val := strings.TrimSpace(string(out))
+	if val == "" {
+		return 0
+	}
+	i, err := strconv.Atoi(val)
+	if err != nil {
+		return 0
+	}
+	return i
+}
+
+// resolveDisplayFromEnv returns the X display string, defaulting to ":1".
+func (s *ApiService) resolveDisplayFromEnv() string {
+	// Prefer KERNEL_IMAGES_API_DISPLAY_NUM, fallback to DISPLAY_NUM, default 1
+	if v := strings.TrimSpace(os.Getenv("KERNEL_IMAGES_API_DISPLAY_NUM")); v != "" {
+		return ":" + v
+	}
+	if v := strings.TrimSpace(os.Getenv("DISPLAY_NUM")); v != "" {
+		return ":" + v
+	}
+	return ":1"
+}
+
+func ptrBool(v bool) *bool { return &v }
+func ptrInt(v int) *int    { return &v }
--- a/server/docs/resolution-change-alternatives.md
+++ b/server/docs/resolution-change-alternatives.md
@@ -1,0 +1,191 @@
+# Alternative Methods for Window Resizing After Resolution Change
+
+This document describes alternative approaches to handle window resizing after changing the display resolution via xrandr, without restarting Chromium.
+
+## Current Implementation
+
+The current implementation in `display.go` restarts Chromium via supervisorctl after a resolution change to ensure the browser window adapts to the new display size. While effective, this approach disrupts the user session.
+
+## Alternative Approaches
+
+### 1. Using xdotool to Resize Windows
+
+**xdotool** is already installed in the image and provides precise window control.
+
+```bash
+# Find and resize all Chromium windows to match new resolution
+xdotool search --class chromium windowsize %@ 1920 1080
+
+# Or move to origin and then resize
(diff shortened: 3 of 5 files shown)


server/openapi.yaml
```diff
@@ -737,6 +737,42 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OkResponse"
+        "409":
+          $ref: "#/components/responses/ConflictError"
+        "400":
+          $ref: "#/components/responses/BadRequestError"
+        "500":
+          $ref: "#/components/responses/InternalError"
```
Sayan-: I think this could make more sense as a `PATCH` SetDisplay` that takes all these params!

server/openapi.yaml
```diff
@@ -737,6 +737,42 @@ paths:
...
+      operationId: displayStatus
+      responses:
+        "200":
+          description: Status
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DisplayStatus"
+        "500":
+          $ref: "#/components/responses/InternalError"
```
Sayan-: I think we should probably just nix this. From an API perspective we should just try to update the display directly over a `get -> set`

server/docs/resolution-change-alternatives.md
```diff
@@ -0,0 +1,191 @@
+# Alternative Methods for Window Resizing After Resolution Change
```
Sayan-: ✂️

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,271 @@
...
+	val := strings.TrimSpace(string(out))
+	if val == "" {
+		return 0
+	}
+	i, err := strconv.Atoi(val)
+	if err != nil {
+		return 0
+	}
+	return i
+}
```
Sayan-: Alternately you could query neko's API since it exposes session information https://github.com/onkernel/neko/blob/aa0487f68ebbff1056d7355ec2da127986e5be5f/server/openapi.yaml#L224-L243

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,627 @@
...
+			return fmt.Errorf("failed to create retry request: %w", err)
+		}
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+
+		resp, err = client.Do(req)
+		if err != nil {
+			return fmt.Errorf("failed to retry Neko API call: %w", err)
+		}
+		defer resp.Body.Close()
```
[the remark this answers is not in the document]
hiroTamada: I guess that is a valid comment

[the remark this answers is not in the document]
hiroTamada: actually, one defer is good enough

images/chromium-headful/Dockerfile
```diff
@@ -129,6 +129,7 @@ RUN set -eux; \
     wget ca-certificates python2 supervisor xclip xdotool \
     pulseaudio dbus-x11 xserver-xorg-video-dummy \
     libcairo2 libxcb1 libxrandr2 libxv1 libopus0 libvpx7 \
+    x11-xserver-utils \
```
hiroTamada, line 132: need this for xrandr

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,627 @@
...
+				Message: fmt.Sprintf("failed to change resolution: %s", err.Error()),
+			},
+		}, nil
+	}
+
+	// Return success with the new dimensions
+	return oapi.PatchDisplay200JSONResponse{
+		Width:  &width,
+		Height: &height,
+	}, nil
```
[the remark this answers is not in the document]
rgarcia: this appears to be correct

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+
+	"github.com/onkernel/kernel-images/server/lib/logger"
+	oapi "github.com/onkernel/kernel-images/server/lib/oapi"
+)
+
+// PatchDisplay updates the display configuration. When require_idle
+// is true (default), it refuses to resize while live view or recording/replay is active.
+// This method automatically detects whether the system is running with Xorg (headful)
+// or Xvfb (headless) and uses the appropriate method to change resolution.
+func (s *ApiService) PatchDisplay(ctx context.Context, req oapi.PatchDisplayRequestObject) (oapi.PatchDisplayResponseObject, error) {
```
rgarcia, line 22: since you're restarting chromium here you probably want to disable scale to zero. See

https://github.com/onkernel/kernel-images/blob/main/server/cmd/api/api/chromium.go#L31-L32

basically this protects against the instance scaling to zero while `PatchDisplay` is doing its thing. We set scale to zero in unikraft to "idle" which will scale an instance to zero on idle connections, which for long-running requests like this is commonly true

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+	log.Info("detected Xorg display (headful mode)")
+	return "xorg"
+}
+
+// restartChromium restarts the Chromium browser via supervisorctl
+func (s *ApiService) restartChromium(ctx context.Context) {
+	log := logger.FromContext(ctx)
+	log.Info("restarting chromium after resolution change")
+
+	restartCmd := []string{"-lc", "supervisorctl restart chromium"}
```
rgarcia: this can be kind of slow to return success, so one thing I did in the custom extensions PR was to put it in the background and consider it complete when the CDP URL is ready (see https://github.com/onkernel/kernel-images/blob/main/server/cmd/api/api/chromium.go#L199-L229)

might be worth it if you're seeing this restart take a long time

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+	log.Info("updating Xvfb resolution requires restart", "width", width, "height", height)
+
+	// Update supervisor config to include environment variables
+	log.Info("updating xvfb supervisor config with new dimensions")
+	removeEnvCmd := []string{"-lc", `sed -i '/^environment=/d' /etc/supervisor/conf.d/services/xvfb.conf`}
+	removeEnvReq := oapi.ProcessExecRequest{Command: "bash", Args: &removeEnvCmd}
+	s.ProcessExec(ctx, oapi.ProcessExecRequestObject{Body: &removeEnvReq})
+
+	// Add the environment line with WIDTH and HEIGHT
+	addEnvCmd := []string{"-lc", fmt.Sprintf(`sed -i '/\[program:xvfb\]/a environment=WIDTH="%d",HEIGHT="%d",DPI="96",DISPLAY=":1"' /etc/supervisor/conf.d/services/xvfb.conf`, width, height)}
```
rgarcia, line 244: this logic is sufficiently intricate where an e2e test might be beneficial. See https://github.com/onkernel/kernel-images/blob/main/server/e2e/e2e_chromium_test.go#L102-L219

server/openapi.yaml
```diff
@@ -1150,6 +1167,54 @@ components:
...
+          minimum: 320
+          description: Display width in pixels
+        height:
+          type: integer
+          minimum: 240
+          description: Display height in pixels
+        refresh_rate:
+          type: integer
+          enum: [60, 30, 25]
+          description: Display refresh rate in Hz. If omitted, uses the highest available rate for the resolution.
```
Sayan-, line 1184: double checking - did we also run numbers on impact of refresh rate in benchmarking? I don't see anything in https://docs.google.com/spreadsheets/d/1VBhIEoNTJoD95gKsalM3XUH2lDnz2XwFlyVwECV-45I/edit?gid=0#gid=0

namely, I'd _think_ higher refresh rates would consume more resources for the gstreamer pipeline backing live views, so I'm curious on the numbers there

server/openapi.yaml
```diff
@@ -1150,6 +1167,54 @@ components:
...
+          description: Number of active Neko viewer sessions.
+        is_recording:
+          type: boolean
+          description: Whether recording is currently active
+        is_replaying:
+          type: boolean
+          description: Whether replay is currently active
+        resizable_now:
+          type: boolean
+          description: True when no blockers are present for resizing
```
Sayan-: do we still need all these bits?

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+	if req.Body.RefreshRate != nil {
+		refreshRate = int(*req.Body.RefreshRate)
+	}
+
+	if width <= 0 || height <= 0 {
+		return oapi.PatchDisplay400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "invalid width/height"}}, nil
+	}
+
+	log.Info("resolution change requested", "width", width, "height", height, "refresh_rate", refreshRate)
```
Sayan-: I think this could be more helpful as a `from` -> `to`!

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+	requireIdle := true
+	if req.Body.RequireIdle != nil {
+		requireIdle = *req.Body.RequireIdle
+	}
+
+	// Check if resize is safe (no active sessions or recordings)
+	if requireIdle {
+		live := s.getActiveNekoSessions(ctx)
+		isRecording := s.anyRecordingActive(ctx)
+		isReplaying := false // replay not currently implemented
```
Sayan-: is this droppable?

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+				},
+			}, nil
+		}
+	}
+
+	// Detect display mode (xorg or xvfb)
+	displayMode := s.detectDisplayMode(ctx)
+
+	// Parse restartChromium flag (default depends on mode)
+	restartChrome := (displayMode == "xvfb") // default true for xvfb, false for xorg
```
Sayan-, line 86: checking my understanding - why don't we restart for headful?

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+				if outputStr == "xvfb" {
+					log.Info("detected Xvfb display (headless mode)")
+					return "xvfb"
+				}
+			}
+		}
+	}
+	log.Info("detected Xorg display (headful mode)")
+	return "xorg"
+}
```
Sayan-, line 146: honestly wonder if this is easier as an env var

server/cmd/api/api/display.go
Sayan-: ==

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+	if err := s.setResolutionViaNeko(ctx, width, height, refreshRate); err != nil {
+		return fmt.Errorf("failed to change resolution via Neko API: %w", err)
+	}
+
+	if restartChrome {
+		s.restartChromium(ctx)
+	}
+
+	return nil
+}
```
Sayan-: do we need this wrapper?

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+
+// isNekoEnabled checks if Neko service is enabled
+func (s *ApiService) isNekoEnabled() bool {
+	return os.Getenv("ENABLE_WEBRTC") == "true"
+}
+
+// getNekoToken obtains a bearer token from Neko API for authentication.
+// It caches the token and reuses it for subsequent requests.
+func (s *ApiService) getNekoToken(ctx context.Context) (string, error) {
```
Sayan-: nit - we should pull the neko port / url from the env instead of hardcoding

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,616 @@
...
+
+	// Check if we have a cached token
+	s.nekoTokenMu.RLock()
+	cachedToken := s.nekoToken
+	s.nekoTokenMu.RUnlock()
+
+	if cachedToken != "" {
+		return cachedToken, nil
+	}
+
```
Sayan-: this check seems redundant. I think we can switch to a vanilla mutex with no real perf changes

server/cmd/api/api/display.go
Sayan-: how long do these tokens last for? do they survive neko restarts? mainly wondering if we could just do it at API start up time so it's always ready but can see there could be problems with that approach

server/openapi.yaml
hiroTamada: on the api side, we are only going to allow these following configurations
```
1024x768x60
1920x1080x60
2560x1440x30
1920 x 1200x60
1440 x 900x30
```

On the images side, I think we can leave it flexible?

server/openapi.yaml
hiroTamada: I guess not... I can just return the current width, height and refresh_rate

server/cmd/api/api/display.go
hiroTamada: This was my thought process.

We can make the images api to be flexible to accommodate both cases where it requires no active sessions or not.

If the request says `requireIdle`, then we reject the request if there are active sessions

Sayan-: ah sorry this was in reference to `isReplaying`. I think it's just a dupe of `isRecording` and it's not implemented

server/cmd/api/api/display.go
hiroTamada: we use this flag to `--start-maximized` to start chromium.

So when x.org expands, the chromium expands. But, we can optionally override this through the request. chromium restarts take time, so we want to avoid it as much as possible

server/cmd/api/api/display.go
hiroTamada: Yeah.... good point. Let me research this. I talked about this with Raf as well. He was more comfortable with this approach

server/cmd/api/api/display.go
hiroTamada: ah I see. Yeah, I will do that. Good thanks!!

server/cmd/api/api/display.go
hiroTamada: I think the token lifetime is technically infinite. But you can manually expire it or rotate it. Maybe it is safer to maintain the current implementation?

server/e2e/e2e_chromium_test.go
```diff
@@ -414,7 +550,8 @@ func runContainer(ctx context.Context, image, name string, env map[string]string
 		"run",
 		"--name", name,
 		"--privileged",
-		"--network=host",
+		"-p", "10001:10001", // API server
+		"-p", "9222:9222", // DevTools proxy
```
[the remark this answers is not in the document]
hiroTamada: I needed this to run the e2e test for some reason. I will take a look

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,429 @@
...
+	height, err := strconv.Atoi(heightStr)
+	if err != nil {
+		log.Error("failed to parse height", "error", err, "value", heightStr)
+		return 1024, 768, 60
+	}
+
+	return width, height, refreshRate
+}
```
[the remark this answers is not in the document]
hiroTamada: that is not a big deal for now since we are setting refreshRate

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,425 @@
...
+
+// getCurrentResolution returns the current display resolution and refresh rate by querying xrandr
+func (s *ApiService) getCurrentResolution(ctx context.Context) (int, int, int) {
+	log := logger.FromContext(ctx)
+	display := s.resolveDisplayFromEnv()
+
+	// Use xrandr to get current resolution
+	cmd := exec.CommandContext(ctx, "bash", "-lc", "xrandr | grep -E '\\*' | awk '{print $1}'")
+	cmd.Env = append(os.Environ(), fmt.Sprintf("DISPLAY=%s", display))
```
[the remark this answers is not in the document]
hiroTamada: that is a valid point. But in isolated micro vm, that is fine for now.

server/e2e/e2e_chromium_test.go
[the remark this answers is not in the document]
hiroTamada: we need this for testing on local machines to work

images/chromium-headful/run-unikernel.sh
```diff
@@ -28,7 +28,32 @@ CHROMIUM_FLAGS="${CHROMIUM_FLAGS:-$chromium_flags_default}"
...
 FLAGS_DIR=".tmp/chromium"
-echo "$CHROMIUM_FLAGS" > "$FLAGS_DIR/flags"
+
+# Convert space-separated flags to JSON array format, handling quoted strings
+# Use eval to properly parse quoted strings (respects shell quoting)
+if [ -n "$CHROMIUM_FLAGS" ]; then
+  eval "FLAGS_ARRAY=($CHROMIUM_FLAGS)"
+else
+  FLAGS_ARRAY=()
+fi
```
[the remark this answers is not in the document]
hiroTamada: that is fine. It is not for production use case

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,426 @@
...
+	log.Info("updating Xvfb resolution requires restart", "width", width, "height", height)
+
+	// Update supervisor config to include environment variables
+	log.Info("updating xvfb supervisor config with new dimensions")
+	removeEnvCmd := []string{"-lc", `sed -i '/^environment=/d' /etc/supervisor/conf.d/services/xvfb.conf`}
+	removeEnvReq := oapi.ProcessExecRequest{Command: "bash", Args: &removeEnvCmd}
+	s.ProcessExec(ctx, oapi.ProcessExecRequestObject{Body: &removeEnvReq})
+
+	// Add the environment line with WIDTH and HEIGHT
+	addEnvCmd := []string{"-lc", fmt.Sprintf(`sed -i '/\[program:xvfb\]/a environment=WIDTH="%d",HEIGHT="%d",DPI="96",DISPLAY=":1"' /etc/supervisor/conf.d/services/xvfb.conf`, width, height)}
```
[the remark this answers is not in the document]
hiroTamada: that is fine. Not user facing

server/e2e/e2e_chromium_test.go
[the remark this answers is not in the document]
hiroTamada: I had to do this to run this on Mac. I will revert if this is not necessary

server/cmd/api/api/display.go
```diff
@@ -0,0 +1,426 @@
...
+// getCurrentResolution returns the current display resolution and refresh rate by querying xrandr
+func (s *ApiService) getCurrentResolution(ctx context.Context) (int, int, int) {
+	log := logger.FromContext(ctx)
+	display := s.resolveDisplayFromEnv()
+
+	// Use xrandr to get current resolution
+	// Note: Using bash -c (not -lc) to avoid login shell overriding DISPLAY env var
+	cmd := exec.CommandContext(ctx, "bash", "-c", "xrandr | grep -E '\\*' | awk '{print $1}'")
+	cmd.Env = append(os.Environ(), fmt.Sprintf("DISPLAY=%s", display))
+
```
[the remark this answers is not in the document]
hiroTamada: not user facing path. This is fine

images/chromium-headful/run-docker.sh
```diff
@@ -22,7 +22,29 @@ CHROMIUM_FLAGS="${CHROMIUM_FLAGS:-$CHROMIUM_FLAGS_DEFAULT}"
...
+    if [ "$FIRST" = true ]; then
+      FLAGS_JSON+="\"$flag\""
+      FIRST=false
+    else
+      FLAGS_JSON+=",\"$flag\""
+    fi
+  fi
+done
+FLAGS_JSON+=']}'
+echo "$FLAGS_JSON" > "$FLAGS_FILE"
```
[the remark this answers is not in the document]
hiroTamada: again, not user facing and this is fine

server/cmd/api/api/api.go
```diff
@@ -47,6 +50,16 @@ func New(recordManager recorder.RecordManager, factory recorder.FFmpegRecorderFa
...
+	// Initialize Neko authenticated client
+	adminPassword := os.Getenv("NEKO_ADMIN_PASSWORD")
+	if adminPassword == "" {
+		adminPassword = "admin" // Default from neko.yaml
+	}
+	nekoAuthClient, err := nekoclient.NewAuthClient("http://127.0.0.1:8080", "admin", adminPassword)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create neko auth client: %w", err)
+	}
+
```
Sayan-: why do we prefer to initial this here instead of passing it in?

server/openapi.yaml
Sayan-: did we also want 10 in here?

server/lib/nekoclient/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+// AuthClient wraps the Neko OpenAPI client and handles authentication automatically.
+// It manages token caching and refresh on 401 responses.
+type AuthClient struct {
+	client   *nekooapi.ClientWithResponses
+	tokenMu  sync.Mutex
+	token    string
+	username string
+	password string
+}
```
Sayan-, line 20: this is way easier to read, thanks for iterating!

server/cmd/api/api/api.go
hiroTamada: no reasons lol

PR #100, opened 2025-12-03
fix: add post-processing step to fix MP4 duration metadata

hiroTamada:
## Summary

Fixes two issues:
1. MP4 files recorded by the recording endpoints don't display the full video length when downloaded (duration extends as videos are played)
2. Concurrent `Stop()` calls could corrupt recordings due to multiple SIGINTs being sent to ffmpeg

### Duration Metadata
The recording uses **fragmented MP4** format (`-movflags +frag_keyframe+empty_moov`) for data safety during recording. This format:
- Writes an empty `moov` atom at the start (without duration information)
- Streams fragments incrementally as it records
- Is great for crash safety, but causes players to calculate duration on playback

### Concurrent Stop Race Condition
When multiple goroutines called `Stop()` concurrently (e.g., from duplicate HTTP requests), each would send SIGINT to ffmpeg. Multiple SIGINTs cause ffmpeg to terminate immediately without flushing buffers, resulting in corrupted/empty MP4 files.

### Post-Processing (Duration Fix)
Add a **post-processing step** that remuxes the fragmented MP4 into a standard MP4 with proper duration metadata after recording stops:
- `finalizeRecording()` method - Remuxes using `ffmpeg -c copy -movflags +faststart`
- No re-encoding (fast operation)
- Moves the `moov` atom to the start with correct duration
- Uses temp file to safely replace the original

### Concurrent-Safe Stop (Race Condition Fix)
Use `sync.Once` to ensure shutdown and finalization only run once:
- `stopOnce` - Ensures shutdown signals are only sent once (prevents duplicate SIGINTs)
- `finalizeOnce` - Ensures finalization only runs once
- Wait for process exit before proceeding to finalization

### API Changes
- `DownloadRecording`: Returns `202 Accepted` with `Retry-After: 5` if recording is being finalized
- `DeleteRecording`: Returns `400` if recording is being finalized; now synchronous instead of async

### Recorder (`server/lib/recorder/ffmpeg.go`)
- Add `finalizeRecording()` to remux fragmented MP4 to standard MP4
- Introduce `ErrRecordingFinalizing` and `finalizing` state
- Gate `Recording()`/`Delete()` to return error while finalizing
- Make shutdown/finalization idempotent via `sync.Once` (`stopOnce`, `finalizeOnce`)
- Wait for process exit before finalizing
- `Stop()` always attempts finalization after graceful shutdown
- `ForceStop()` attempts finalization with warning-only on failure

### API (`server/cmd/api/api/api.go`)
- `DownloadRecording`: Return `202` with `Retry-After: 5` if finalizing
- `DeleteRecording`: Synchronous delete; return `400` if finalizing

### Tests (`server/lib/recorder/ffmeg_test.go`)
- Set explicit `outputPath` and adjust expectations for new finalization behavior

## Testing

Tested with concurrent stop script:
go run ./scripts/concurrent_stop_test/main.go -url http://localhost:444 -concurrency 5 -iterations 10Results: **10/10 passed** with 5 concurrent stop calls per iteration.

---

> [!NOTE]
> Adds FFmpeg post-processing to finalize MP4 duration metadata and singleflight-guards Stop/finalize; API now waits for finalization on download and returns 409 on delete during finalization.
> 
> - **Recorder (FFmpeg)**:
>   - Add `finalizeRecording()` remux step (faststart) and finalize automatically on natural exits; expose `WaitForFinalization`.
>   - Introduce `ErrRecordingFinalizing`; gate `Recording()`/`Delete()` until finalization completes.
>   - Use `singleflight` to dedupe concurrent `Stop()` and finalization; `Stop()` always attempts finalize; `ForceStop()` tries finalize but warns on failure.
>   - Update `waitForCommand()` to trigger finalization; `FFmpegManager.StopAll()` now calls `Stop()` for all.
> - **API**:
>   - `StopRecording`: always invoke `Stop()` even if not recording.
>   - `DownloadRecording`: if finalizing, wait and then serve; still returns `202` when in-progress and tiny file.
>   - `DeleteRecording`: make synchronous; return `409 Conflict` when finalizing; log improvements.
> - **OpenAPI/Client**:
>   - Add `409 Conflict` response to `DELETE /recording/delete`; regenerate `oapi` client/server stubs and embedded swagger.
> - **Tests**:
>   - Update recorder tests to set explicit `outputPath` and assert new finalization/force-stop behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d3e043f435d3c75cc4008c35d79284eb417d5390. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/server/lib/recorder/ffmpeg.go
+++ b/server/lib/recorder/ffmpeg.go
@@ -282,12 +282,19 @@ func (fr *FFmpegRecorder) finalizeRecording(ctx context.Context) error {
 	_, err, _ := fr.flight.Do("finalize", func() (any, error) {
 		log := logger.FromContext(ctx)
 
-		// Check if already complete (handles callers after previous singleflight completed)
 		fr.mu.Lock()
+		// Check if already complete (handles callers after previous singleflight completed)
 		if fr.finalizeComplete {
 			result := fr.finalizeResultErr
 			fr.mu.Unlock()
 			return nil, result
+		}
+		// Guard: finalization requires the recording to have exited.
+		// Finalizing an active recording would corrupt it (remux reads partial file,
+		// then os.Rename orphans the inode ffmpeg is writing to).
+		if fr.exitCode < exitCodeProcessDoneMinValue {
+			fr.mu.Unlock()
+			return nil, fmt.Errorf("cannot finalize: recording is still in progress")
 		}
 		outputPath := fr.outputPath
 		binaryPath := fr.binaryPath


server/lib/recorder/ffmpeg.go
```diff
@@ -200,23 +200,105 @@ func (fr *FFmpegRecorder) Stop(ctx context.Context) error {
...
+		// Clean up temp file on error
+		os.Remove(tempPath)
+		return fmt.Errorf("failed to finalize recording: %w", err)
+	}
+
+	// Replace original with finalized version
+	if err := os.Rename(tempPath, outputPath); err != nil {
+		os.Remove(tempPath)
+		return fmt.Errorf("failed to replace recording with finalized version: %w", err)
+	}
```
Sayan-: does this need to be done under a lock to be threadsafe?

Sayan- (commented):
I think in general this looks good! I had a few questions:

1. How much time does the copy add for longer recordings? For example if someone takes a 15+ minute replay how much extra time do we add?
2. Since a replay will temp double the amt of disk space it uses are there any guards we need to think about adding?
3. During finalization what happens if someone is currently streaming one of the replays? Will the stream break / be corrupted?

server/lib/recorder/ffmpeg.go
hiroTamada: good point. Will add it

server/lib/recorder/ffmpeg.go
```diff
@@ -195,28 +210,137 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+	fr.finalizeOnce.Do(func() {
+		fr.finalizeErr = fr.finalizeRecording(context.WithoutCancel(ctx))
+	})
+	if fr.finalizeErr != nil {
+		// Log but don't fail - the recording may still be partially usable
+		log.Warn("failed to finalize force-stopped recording", "err", fr.finalizeErr)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
hiroTamada: that is kind of a point of `ForceStop`. It is by designed

server/lib/recorder/ffmpeg.go
```diff
@@ -195,28 +210,137 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+
+	// Remux the fragmented MP4 to add proper duration metadata.
+	// We proceed with finalization even if ffmpeg exited with a non-zero code (e.g., 255 from SIGINT)
+	// because the recording file is still valid and needs proper duration metadata.
+	// Use WithoutCancel to preserve logger/values but ignore caller cancellation (matching shutdown design)
+	// Use sync.Once to ensure finalization only runs once even with concurrent Stop() calls.
+	fr.finalizeOnce.Do(func() {
+		fr.finalizeErr = fr.finalizeRecording(context.WithoutCancel(ctx))
+	})
+	return fr.finalizeErr
```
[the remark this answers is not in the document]
hiroTamada: this is a good comment

server/cmd/api/api/api.go
```diff
@@ -129,10 +129,10 @@ func (s *ApiService) StopRecording(ctx context.Context, req oapi.StopRecordingRe
 	if !exists {
 		log.Error("attempted to stop recording when none is active", "recorder_id", recorderID)
 		return oapi.StopRecording400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "no active recording to stop"}}, nil
-	} else if !rec.IsRecording(ctx) {
-		log.Warn("recording already stopped", "recorder_id", recorderID)
-		return oapi.StopRecording200Response{}, nil
 	}
+	// Always call Stop() even if IsRecording() is false to ensure finalization runs.
+	// Recordings that exit naturally (max duration, max file size, etc.) need Stop()
+	// to trigger finalization which adds proper duration metadata to the MP4.
```
Sayan-: Hmm - I'd want natural exits to also finalize immediately. What's preventing us from doing this?

server/cmd/api/api/api.go
```diff
@@ -182,12 +182,21 @@ func (s *ApiService) DownloadRecording(ctx context.Context, req oapi.DownloadRec
 
 	out, meta, err := rec.Recording(ctx)
 	if err != nil {
+		if errors.Is(err, recorder.ErrRecordingFinalizing) {
+			log.Info("recording is being finalized, client should retry", "recorder_id", recorderID)
+			return oapi.DownloadRecording202Response{
+				Headers: oapi.DownloadRecording202ResponseHeaders{
+					RetryAfter: 5, // finalization is typically fast
+				},
+			}, nil
```
Sayan-: should we just wait for finalization here instead of telling the client to retry?

server/lib/recorder/ffmpeg.go
```diff
@@ -195,28 +210,137 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
-	// shutdown process from any inbound context
-	err := fr.shutdownInPhases(context.Background(), []shutdownPhase{
-		{"wake_and_interrupt", []syscall.Signal{syscall.SIGINT}, time.Minute, "graceful stop"},
-		{"terminate", []syscall.Signal{syscall.SIGTERM}, 2 * time.Second, "forceful termination"},
-		{"kill", []syscall.Signal{syscall.SIGKILL}, 100 * time.Millisecond, "immediate kill"},
+
+	// Use sync.Once to ensure shutdown signals are only sent once.
+	// This prevents multiple SIGINTs from being sent to ffmpeg when Stop() is called
+	// concurrently, which would cause immediate termination without buffer flushing.
+	fr.stopOnce.Do(func() {
```
Sayan-: take it or leave it - I think this is all semantically correct but using https://pkg.go.dev/golang.org/x/sync/singleflight#Group.Do is actually easier to reason about (e.g. it coordinates duplicate callers and preserves the output of the first run). especially in the context of our typical flow, which is `stop -> download -> save to somewhere durable`

server/lib/recorder/ffmpeg.go
```diff
@@ -195,28 +210,137 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+			return fr.stopErr
+		}
+	}
+
+	// Remux the fragmented MP4 to add proper duration metadata.
+	// We proceed with finalization even if ffmpeg exited with a non-zero code (e.g., 255 from SIGINT)
+	// because the recording file is still valid and needs proper duration metadata.
+	// Use WithoutCancel to preserve logger/values but ignore caller cancellation (matching shutdown design)
+	// Use sync.Once to ensure finalization only runs once even with concurrent Stop() calls.
+	fr.finalizeOnce.Do(func() {
```
Sayan-: should we be `finalizeOnce`'ing from inside of `finalizeRecording`? why or why not?

Sayan- (commented):
awesome stuff! had a few small q's!

server/lib/recorder/ffmpeg.go
hiroTamada: Inside the function is cleaner. At the caller level gives more control (e.g., the completion-state caching we added). For this case, inside is probably simpler since all callers want the same behavior.

I will refactor it.

server/cmd/api/api/api.go
```diff
@@ -182,12 +182,34 @@ func (s *ApiService) DownloadRecording(ctx context.Context, req oapi.DownloadRec
...
+		if errors.Is(err, recorder.ErrRecordingFinalizing) {
+			// Wait for finalization to complete instead of asking client to retry
+			log.Info("waiting for recording finalization", "recorder_id", recorderID)
+			ffmpegRec, ok := rec.(*recorder.FFmpegRecorder)
+			if !ok {
+				log.Error("failed to cast recorder to FFmpegRecorder", "recorder_id", recorderID)
+				return oapi.DownloadRecording500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "internal error"}}, nil
+			}
+			// WaitForFinalization blocks until finalization completes and returns the result
+			if finalizeErr := ffmpegRec.WaitForFinalization(ctx); finalizeErr != nil {
```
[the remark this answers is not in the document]
hiroTamada: Finalization is not too critical. I think this is fine

server/lib/recorder/ffmpeg.go
```diff
@@ -48,6 +53,16 @@ type FFmpegRecorder struct {
...
 	deleted    bool
 	stz        *scaletozero.Oncer
+
+	// stopFlight coordinates concurrent Stop() calls to prevent multiple SIGINTs
+	// from being sent to ffmpeg, which causes immediate abort without proper file closure.
+	stopFlight singleflight.Group
+
+	// finalizeFlight coordinates finalization across Stop(), ForceStop(), and waitForCommand().
+	// All callers block until finalization completes and receive the same result.
+	finalizeFlight    singleflight.Group
```
Sayan-: nit - you can use the same singleflight group for everything. each `Do` has a unique namespace, which we could use instead to differentiate between `stop` and `finalize` like you're already doing

server/lib/recorder/ffmpeg.go
```diff
@@ -195,30 +210,157 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+		if errMsg == "failed to shutdown ffmpeg" || errMsg == "no recording to stop" {
+			return shutdownErr
+		}
+	}
+
+	// Wait for the process to exit (in case another goroutine is handling shutdown).
+	// This ensures we don't proceed to finalization while ffmpeg is still running.
+	if fr.exited != nil {
+		<-fr.exited
+	}
```
Sayan-: is this case still necessary since the goroutines are coordinated with singleflight now?

server/lib/recorder/ffmpeg.go
```diff
@@ -195,30 +210,157 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+		fr.finalizeResultErr = result
+		fr.mu.Unlock()
+
+		return nil, result
+	})
 	return err
 }
 
+// doFinalize performs the actual remuxing work.
+func (fr *FFmpegRecorder) doFinalize(ctx context.Context) error {
```
Sayan-: I think the llm really liked factoring out the call inside of singleflight but since this function explicitly shouldn't be called outside of `finalizeRecording` for a number of reasons (correctness, reliability) I think inlining the definition or keeping all of the logic inside of the `Do` is actually more maintainable

server/lib/recorder/ffmpeg.go
```diff
@@ -195,30 +210,157 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+	return nil
+}
+
+// finalizeRecording remuxes the fragmented MP4 to create a standard MP4 with
+// proper duration metadata in the moov atom. This is necessary because fragmented
+// MP4 (used for data safety during recording) doesn't include duration in the header.
+//
+// This method is safe for concurrent calls - it uses singleflight internally to ensure
+// finalization runs exactly once, with all callers receiving the same result.
+func (fr *FFmpegRecorder) finalizeRecording(ctx context.Context) error {
```
Sayan-, line 281: I see a number of calls to this function passing in a `context.WithoutCancel`. consider inverting it where we internally call `ExecContext(context.WithoutCancel(...))` if no cancellation is required for correctness / prevent data loss or corruption

server/cmd/api/api/api.go
```diff
@@ -224,15 +246,18 @@ func (s *ApiService) DeleteRecording(ctx context.Context, req oapi.DeleteRecordi
 		return oapi.DeleteRecording400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "recording must be stopped first"}}, nil
 	}
 
-	// fine to do this async
-	go func() {
-		if err := rec.Delete(context.Background()); err != nil && !errors.Is(err, os.ErrNotExist) {
+	if err := rec.Delete(ctx); err != nil {
+		if errors.Is(err, recorder.ErrRecordingFinalizing) {
+			log.Info("recording is being finalized, client should retry", "recorder_id", recorderID)
+			return oapi.DeleteRecording400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: "recording is being finalized, please retry in a few seconds"}}, nil
```
Sayan-: would a 409 be more clear here? https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#:~:text=any%20later%20time.%22-,409%20Conflict,-Indicates%20that%20the

server/lib/recorder/ffmpeg.go
```diff
@@ -505,11 +670,11 @@ func (fm *FFmpegManager) StopAll(ctx context.Context) error {
 
 	var errs []error
 	for id, recorder := range fm.recorders {
-		if recorder.IsRecording(ctx) {
-			if err := recorder.Stop(ctx); err != nil {
-				errs = append(errs, fmt.Errorf("failed to stop recorder '%s': %w", id, err))
-				log.Error("failed to stop recorder during shutdown", "id", id, "err", err)
-			}
+		// Always call Stop() to ensure finalization runs, even for recordings that
+		// exited naturally. Stop() handles already-exited processes gracefully.
```
Sayan-: is this comment still up to date?

server/lib/recorder/ffmpeg.go
hiroTamada: I see. That makes sense

server/lib/recorder/ffmpeg.go
hiroTamada: good catch. This is not update-to-date

server/lib/recorder/ffmpeg.go
```diff
@@ -195,27 +207,142 @@ func (fr *FFmpegRecorder) Start(ctx context.Context) error {
...
+	// because the recording file is still valid and needs proper duration metadata.
+	return fr.finalizeRecording(ctx)
+}
+
+// WaitForFinalization blocks until finalization completes and returns the result.
+// If finalization hasn't started, it will be triggered. If already complete, returns
+// the cached result immediately. This is useful for callers like the download handler
+// that need to wait for finalization before accessing the recording.
+func (fr *FFmpegRecorder) WaitForFinalization(ctx context.Context) error {
+	return fr.finalizeRecording(ctx)
```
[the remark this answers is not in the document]
hiroTamada: I can add a safe guard, but it is probably safe to assume that `WaitForFinalization` is called after recoding is finished

server/cmd/api/api/api.go
```diff
@@ -182,12 +182,34 @@ func (s *ApiService) DownloadRecording(ctx context.Context, req oapi.DownloadRec
...
+			if finalizeErr := ffmpegRec.WaitForFinalization(ctx); finalizeErr != nil {
+				log.Error("finalization failed", "err", finalizeErr, "recorder_id", recorderID)
+				return oapi.DownloadRecording500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "failed to finalize recording"}}, nil
+			}
+			// Finalization complete, retry getting the recording
+			out, meta, err = rec.Recording(ctx)
+			if err != nil {
+				log.Error("failed to get recording after finalization", "err", err, "recorder_id", recorderID)
+				return oapi.DownloadRecording500JSONResponse{InternalErrorJSONResponse: oapi.InternalErrorJSONResponse{Message: "failed to get recording"}}, nil
+			}
```
Sayan-, line 203: non-blocking: the fact that we're doing the equivalent of a `Get` -> `Wait` -> `Get` again outside of the interface and relying on casting (`ffmpegRec, ok := rec.(*recorder.FFmpegRecorder)`) indicates we've broken the abstraction. consider making `Recording` just wait for finalization instead of making that a responsibility of the caller

server/lib/recorder/ffmpeg.go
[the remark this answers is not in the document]
Sayan-: I think it's also confusing that `WaitForFinalization` can trigger mutation.

hiroTamada:
> I don't want to block this going out but I think in general the `FFmpegRecorder` doesn't need to expose the semantics of finalization to callers. It increases complexity in the API endpoints rather than isolating it into a single entity.
> 
> Particularly:
> 
> * For Delete, either block or return a generic temporary/busy error the API maps to 409 (avoid finalize-specific sentinel in the public surface)
> * Make Recording block; don’t expose ErrRecordingFinalizing there.

Hmmm, I do agree with analysis. If you are ok with it, should we have another review cycle?

Sayan-:
Yeah I think it would be good to do in a quick follow up!

PR #110, opened 2025-12-19
[kernel 726] fix web bot auth extension

archandatta:
> [!NOTE]
> Introduces enterprise policy-based installation for extensions and serves required files, improving reliability for extensions needing `webRequest`/`webRequestBlocking`.
> 
> - In `UploadExtensionsAndRestart`, detect policy-required extensions, extract Chrome extension ID from `update.xml`, validate presence of `update.xml` and `.crx`, update enterprise policy, and only add `--load-extension` flags for non-policy extensions
> - Extend policy model: add `ExtensionInstallForcelist`, use `update_url` instead of local `path`, dedupe forcelist entries, and add `ExtractExtensionIDFromUpdateXML` with ID validation
> - Serve `http://.../extensions/*` from `/home/kernel/extensions` so Chrome can fetch `update.xml` and `.crx`
> - E2E: add `TestWebBotAuthInstallation` verifying `ExtensionInstallForcelist` and extension directory
> - Docker: make `/chromium/flags` bind mount writable
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 46a6e1e38d1272ece019aa7574522dd48bbea542. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/server/cmd/api/main.go
+++ b/server/cmd/api/main.go
@@ -133,32 +133,6 @@ func main() {
 		fs.ServeHTTP(w, r)
 	})
 
-	// Serve CRX files at root for Chrome enterprise policy
-	// This allows simple codebase URLs like http://host:port/extension-name.crx
-	r.Get("/{filename}.crx", func(w http.ResponseWriter, r *http.Request) {
-		// Extract the filename from the URL path
-		filename := chi.URLParam(r, "filename") + ".crx"
-
-		// Search for the CRX file in all extension directories
-		entries, err := os.ReadDir(extensionsDir)
-		if err != nil {
-			http.Error(w, "extensions directory not found", http.StatusNotFound)
-			return
-		}
-
-		for _, entry := range entries {
-			if entry.IsDir() {
-				crxPath := fmt.Sprintf("%s/%s/%s", extensionsDir, entry.Name(), filename)
-				if _, err := os.Stat(crxPath); err == nil {
-					http.ServeFile(w, r, crxPath)
-					return
-				}
-			}
-		}
-
-		http.Error(w, "crx file not found", http.StatusNotFound)
-	})
-
 	srv := &http.Server{
 		Addr:    fmt.Sprintf(":%d", config.Port),
 		Handler: r,


server/cmd/api/api/chromium.go
```diff
@@ -157,49 +158,106 @@ func (s *ApiService) UploadExtensionsAndRestart(ctx context.Context, request oap
...
 		requiresEntPolicy, err := s.policy.RequiresEnterprisePolicy(manifestPath)
 		if err != nil {
-			log.Warn("failed to read manifest for policy check", "error", err, "extension", p.name)
+			log.Warn("failed to read manifest for policy check", "error", err, "extension", extensionName)
 			// Continue with requiresEntPolicy = false
 		}
 
+		// Try to extract Chrome extension ID from update.xml
+		chromeExtensionID := extensionName
+		extractionErr := error(nil)
```
rgarcia: nit: `var extractionErr error` is more idiomatic

server/cmd/api/main.go
```diff
@@ -124,6 +124,64 @@ func main() {
...
+	extensionsDir := "/home/kernel/extensions"
+	r.Get("/extensions/*", func(w http.ResponseWriter, r *http.Request) {
+		// Serve files from /home/kernel/extensions/
+		fs := http.StripPrefix("/extensions/", http.FileServer(http.Dir(extensionsDir)))
+		fs.ServeHTTP(w, r)
+	})
+
+	// Serve update.xml at root for Chrome enterprise policy
+	// This serves the first update.xml found in any extension directory
+	r.Get("/update.xml", func(w http.ResponseWriter, r *http.Request) {
```
rgarcia: this serves the first `update.xml` found based on directory iteration order, which isn't deterministic. if there are multiple extensions with `update.xml` files, this could return the wrong one. is this route needed given `/extensions/{name}/update.xml` already works via the wildcard route above?

server/lib/policy/policy.go
```diff
@@ -113,22 +125,46 @@ func (p *Policy) AddExtension(extensionID, extensionPath string, requiresEnterpr
...
+
+		// Remove any existing entries with the same extension ID (different URLs)
+		if policy.ExtensionInstallForcelist == nil {
+			policy.ExtensionInstallForcelist = []string{}
+		}
+
+		// Filter out entries that start with the same extension ID
+		extensionIDPrefix := chromeExtensionID + ";"
+		policy.ExtensionInstallForcelist = slices.DeleteFunc(policy.ExtensionInstallForcelist, func(entry string) bool {
+			return len(entry) >= len(extensionIDPrefix) && entry[:len(extensionIDPrefix)] == extensionIDPrefix
```
rgarcia: nit: `strings.HasPrefix(entry, extensionIDPrefix)` would be cleaner here

server/lib/policy/policy.go
```diff
@@ -168,3 +204,45 @@ func (p *Policy) RequiresEnterprisePolicy(manifestPath string) (bool, error) {
...
+	if len(manifest.Apps) == 0 {
+		return "", fmt.Errorf("no <app> element found in update.xml")
+	}
+
+	appID := manifest.Apps[0].AppID
+	if appID == "" {
+		return "", fmt.Errorf("appid attribute is empty in update.xml")
+	}
+
+	// Validate extension ID format: Chrome extension IDs are 32 lowercase hex characters
```
rgarcia: nit: comment says "hex" but Chrome extension IDs use a-p (a base16 variant), not 0-9a-f. maybe "32 lowercase a-p characters" to match the regex

server/lib/policy/policy.go
```diff
@@ -168,3 +204,45 @@ func (p *Policy) RequiresEnterprisePolicy(manifestPath string) (bool, error) {
...
+	// This prevents injection attacks via semicolons or other special characters
+	if !regexp.MustCompile(`^[a-p]{32}$`).MatchString(appID) {
```
rgarcia: nit: this compiles the regex on every call. could move to a package-level `var extensionIDRegex = regexp.MustCompile(...)`

server/e2e/e2e_chromium_test.go
```diff
@@ -804,115 +804,3 @@ func listCDPTargets(ctx context.Context) ([]map[string]interface{}, error) {
 
 	return targets, nil
 }
-
-func TestWebBotAuthInstallation(t *testing.T) {
```
rgarcia, line 808: this test was verifying the web-bot-auth policy installation flow. rather than deleting it, could we update it to test the new behavior? the new flow (requiring `update.xml` + `.crx`, `ExtensionInstallForcelist`, `update_url` instead of `path`) seems important enough to have e2e coverage.

rgarcia (changes requested):
Good progress on the Chrome enterprise policy implementation. The core logic for separating policy vs non-policy extensions looks solid.

Main concern: the deleted `TestWebBotAuthInstallation` test should be updated rather than removed - the new flow has significant behavior changes that warrant e2e coverage.

A few minor nits on style/perf, and a question about whether the `/update.xml` root route is necessary.

rgarcia (approved):
Changes since last review look good:

- All nits addressed (idiomatic error declaration, `strings.HasPrefix`, package-level regex)
- Removed non-deterministic `/update.xml` and `/{filename}.crx` routes
- Re-added `TestWebBotAuthInstallation` with proper update.xml and .crx files
- `RuntimeAllowedHosts` removal is correct - matches upstream web-bot-auth reference and force-installed extensions get manifest permissions automatically

Ship it 🚀

PR #113, opened 2026-01-07
perf: add persistent playwright executor daemon

rgarcia:
## Summary

Replace per-request `tsx` subprocess spawning with a persistent Node.js daemon that maintains a warm CDP connection to the browser.

**Performance improvement: ~94% faster for simple operations**
- Before: ~750ms per request (tsx startup + module load + CDP connect)
- After: ~45ms per request (reuses daemon + warm CDP connection)

### Changes

- Add `playwright-daemon.ts`: persistent daemon that listens on Unix socket, maintains warm CDP connection, uses esbuild for fast TypeScript transformation
- Update `playwright.go`: communicate with daemon via Unix socket instead of spawning tsx subprocess; lazy-start daemon on first request
- Update Dockerfiles: install esbuild, pre-compile daemon to JavaScript during build

### Architecture

```
HTTP Request → Go API → Unix Socket → Playwright Daemon → CDP → Chromium
                              ↑
                    (persistent, warm connection)
```

## Test plan

- [x] Build headless Docker image
- [x] Run performance benchmarks comparing before/after
- [x] Test TypeScript syntax support (type annotations, interfaces, async/await)
- [x] Verify existing Go tests pass
- [ ] Manual testing in production environment

---

> [!NOTE]
> Introduces a persistent Playwright execution daemon and switches the API to use it over a Unix socket, removing per-request tsx subprocesses.
> 
> - **New daemon**: Adds `server/runtime/playwright-daemon.ts` (compiled to `/usr/local/lib/playwright-daemon.js`) that listens on `/tmp/playwright-daemon.sock`, keeps a warm CDP connection, transforms TS via `esbuild`, and reconnects on browser restarts.
> - **API changes**: `api/playwright.go` now lazy-starts the daemon, communicates via Unix socket with request/response JSON, and returns structured errors; `ApiService` gains fields for daemon process control.
> - **Dockerfiles**: Install `esbuild`, remove `tsx` from globals, and compile+embed the daemon in both headful/headless images; update neko base image reference.
> - **Tests**: Move Playwright execute test into `e2e_playwright_test.go` and add a recovery test that restarts Chromium and verifies daemon reconnection.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4be529bac1625ee9cac348fdb692cb57cd7ae5e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/images/chromium-headful/Dockerfile
+++ b/images/chromium-headful/Dockerfile
@@ -146,7 +146,7 @@ RUN --mount=type=cache,target=/tmp/cache/ffmpeg,sharing=locked,id=$CACHEIDPREFIX
     rm -rf /tmp/ffmpeg*
 EOT
 
-FROM ghcr.io/onkernel/neko/base:3.0.8-v1.3.0 AS neko
+FROM ghcr.io/kernel/neko/base:3.0.8-v1.3.0 AS neko
 # ^--- now has event.SYSTEM_PONG with legacy support to keepalive
 FROM node:22-bullseye-slim AS node-22
 FROM docker.io/ubuntu:22.04


server/runtime/playwright-daemon.ts
```diff
@@ -0,0 +1,266 @@
+/**
+ * Persistent Playwright Executor Daemon
```
ulziibay-kernel, line 2: if i want to add tracing functionality a first class citizen, is this a place to add that given the architecture?

rgarcia: potentially! Currently there are two main ways users can execute playwright code against a Kernel browser:
(1) spin up a kernel browser, and then use the CDP url from a playwright script. Here we have no control over playwright tracing config since it's user code
(2) use the playwright execute API (https://www.kernel.sh/docs/browsers/playwright-execution) which is what this PR modifies. This is an API that takes typescript an executes it against the browser

i'd say most people use (1) right now so if we added a parameter to enable tracing in this endpoint it might not be useful in practice

server/runtime/playwright-daemon.ts
```diff
@@ -0,0 +1,266 @@
...
+
+import { createServer, Socket } from 'net';
+import { unlinkSync, existsSync } from 'fs';
+import { transform } from 'esbuild';
+import { chromium as chromiumPW, Browser } from 'playwright-core';
+import { chromium as chromiumPR } from 'patchright';
+
+const SOCKET_PATH = process.env.PLAYWRIGHT_DAEMON_SOCKET || '/tmp/playwright-daemon.sock';
+const CDP_ENDPOINT = process.env.CDP_ENDPOINT || 'ws://127.0.0.1:9222';
+const USE_PATCHRIGHT = process.env.PLAYWRIGHT_ENGINE !== 'playwright-core';
```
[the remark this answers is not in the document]
rgarcia: intentional

server/runtime/playwright-daemon.ts
```diff
@@ -0,0 +1,266 @@
...
+    const userFunction = new AsyncFunction('page', 'context', 'browser', jsCode);
+
+    const timeoutPromise = new Promise<never>((_, reject) => {
+      setTimeout(() => reject(new Error(`Execution timed out after ${timeout_ms}ms`)), timeout_ms);
+    });
+
+    const result = await Promise.race([
+      userFunction(page, context, browserInstance),
+      timeoutPromise,
+    ]);
```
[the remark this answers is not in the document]
rgarcia: Acknowledged. This is an inherent tradeoff of the persistent daemon architecture. A few mitigating factors:
 
1. The Go API serializes all requests with a mutex, so concurrent execution isn't possible
2. Timed-out code is typically blocked on I/O (awaiting network), not CPU-bound, so it won't starve subsequent requests
3. The practical risk of page state corruption is low - most timeouts occur during navigation/network issues where the page is already in a bad state
 
True isolation would require per-request subprocesses (the old model) or worker threads with their own CDP connections, both of which sacrifice the ~94% latency improvement this PR provides.
 
We accept this tradeoff for the performance gains. If it becomes a problem in practice, we can revisit with a "fresh page after timeout" mitigation.

server/runtime/playwright-daemon.ts
```diff
@@ -0,0 +1,266 @@
+/**
```
Sayan-, line 1: should we just manage the daemon via supervisord instead? that way we can also automatically resume from failures, etc. wdyt?

server/cmd/api/api/playwright.go
```diff
@@ -29,107 +144,42 @@ func (s *ApiService) ExecutePlaywrightCode(ctx context.Context, request oapi.Exe
...
-			return oapi.ExecutePlaywrightCode200JSONResponse{
-				Success: success,
-				Error:   &errorMsg,
-				Stderr:  &stderr,
-			}, nil
-		}
-
-		// If we can't parse the output, return a generic error
-		success := false
 		errorMsg := fmt.Sprintf("execution failed: %v", err)
```
Sayan-, line 164: might be worth distinguishing daemon vs code execution errors to make debugging easier

Sayan- (commented):
sweet! qq on daemon management, lgtm otherwise

server/runtime/playwright-daemon.ts
rgarcia: since CDP connections present some bot detection risk i figured we should only run the daemon if the user is using this playwright endpoint, so that's why it's not running by default

Sayan-: sg! exciting stuff!

PR #169, opened 2026-03-02
Humanize all computer endpoints

ulziibay-kernel:
This is a plan to implement human like interactions for all Computer Controls. https://www.kernel.sh/docs/browsers/computer-controls#computer-controls. 
https://github.com/kernel/kernel-images/pull/148 added human like mouse movements. However, there are other movements like clicking, scrolling, typing, pressing that could use human like improvements.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change that adds an implementation plan; no runtime code, APIs, or generated artifacts are modified in this PR.
> 
> **Overview**
> Adds `plans/humanize-computer-endpoints.md`, a detailed performance-focused design plan for humanizing all computer interaction endpoints (click, type, press key, scroll, drag) while minimizing `xdotool` process spawns.
> 
> The document proposes new `smooth` flags and typing options (e.g., optional typo injection), plus a small shared `humanize` helper library and related OpenAPI/type regeneration steps, but these are **not implemented here**.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8fb87c446743a3702c61cf82cbdb0cefdaf7211. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


plans/humanize-computer-endpoints.md
```diff
@@ -0,0 +1,210 @@
...
+
+**API change:** Add `smooth: boolean` (default `true`) to `ClickMouseRequest`.
+
+---
+
+## 2. Type Text -- Chunked Type with Inter-Word Pauses
+
+**Cost:** 1 xdotool call (same as current). Pre-computation: O(words) random samples.
+
+**Algorithm:** Instead of per-character keysym mapping (which is complex and fragile for Unicode), split text by whitespace/punctuation into chunks and chain `xdotool type --delay <intra> "chunk" sleep <inter>` commands.
```
rgarcia: i don’t think this works with `xdotool` command chaining as written. `type` consumes the rest of the argv, so `type ... sleep ... type ...` won’t execute as separate commands in one invocation. if we want chunked typing with pauses, i think this needs `xdotool` script mode or a different execution strategy.

plans/humanize-computer-endpoints.md
```diff
@@ -0,0 +1,210 @@
...
+
+- **Intra-word delay**: Per-chunk, pick `rand.Intn(70) + 50` -> [50, 120]ms. Varies per chunk to simulate burst-pause rhythm.
+- **Inter-word pause**: Between chunks, insert `sleep` with `UniformJitter(rng, 140, 60, 60)` -> [80, 200]ms. Longer pauses at sentence boundaries (after `.!?`): multiply by 1.5x.
+- **No bigram tables**: The per-word delay variation is sufficient for convincing humanization. Bigram-level precision adds complexity with diminishing returns for bot detection evasion.
+
+**Single xdotool call example:**
+
+```
+xdotool type --delay 80 -- "Hello" sleep 0.150 type --delay 65 -- " world" sleep 0.300 type --delay 95 -- ". How" sleep 0.120 type --delay 70 -- " are" sleep 0.140 type --delay 85 -- " you?"
```
rgarcia: small algo nit: this chunking puts the pause before the delimiter instead of after it, since the separator is attached to the next chunk (`" world"`, `". How"`). that means we wait before typing the space or period, which is the opposite of the “pause after a word/sentence” rhythm we probably want. if we keep this approach, i’d split so whitespace/punctuation stays with the previous chunk.

plans/humanize-computer-endpoints.md
```diff
@@ -0,0 +1,210 @@
...
+
+---
+
+## 4. Scroll -- Eased Tick Intervals in One Call
+
+**Cost:** 1 xdotool call (same as current). Pre-computation: O(ticks) easing function evaluations (3 multiplies each).
+
+**Algorithm:** Replace `click --repeat N --delay 0 <btn>` with N individual `click <btn>` commands separated by pre-computed `sleep` values following a **smoothstep easing curve**.
+
+- **Easing**: `SmoothStepDelay(i, N, slowMs=80, fastMs=15)` for each tick i. The smoothstep `3t^2 - 2t^3` creates natural momentum: slow start, fast middle, slow end.
```
rgarcia: the timing here feels calibrated too high for this API’s scroll units. `delta_x` / `delta_y` are logical wheel ticks, not pixels, so `slowMs=80` / `fastMs=15` makes even moderate scrolls take a pretty long time. for example, a `delta_y=10` scroll is already on the order of ~0.5s with this profile. i’d either use much smaller per-tick delays or derive delays from a bounded total-duration target so larger wheel counts don’t block input for so long.

plans/humanize-computer-endpoints.md
ulziibay-kernel: Good catch -- you're right, `xdotool type` consumes the rest of argv so it can't be chained with `sleep` in a single invocation.

Updated the plan to use **O(words) separate `xdotool type` calls with Go-side `sleepWithContext` pauses** between them. This follows the same pattern as `doMoveMouseSmooth` (O(n) calls with Go-side sleeps). For a typical sentence of 5-15 words, the inter-word pauses (80-300ms) dwarf the ~1-2ms fork+exec overhead, so the performance impact is negligible.

See 627054d.

plans/humanize-computer-endpoints.md
ulziibay-kernel: Good call. Updated the chunking so trailing delimiters (spaces, punctuation) stay with the preceding chunk:

`"Hello world. How are you?"` → `["Hello ", "world. ", "How ", "are ", "you?"]`

This way the pause happens *after* the user types the space/period, matching natural typing rhythm. Sentence boundary detection (`chunk ends with .!?`) still works since the punctuation is now at the end of the chunk.

See 627054d.

plans/humanize-computer-endpoints.md
ulziibay-kernel: Agreed, the per-tick delays were way too high for logical wheel ticks. Updated to use a **bounded total duration** approach:

- Default `totalMs = 200` regardless of tick count
- Per-tick delay = `totalMs / N`, then shaped by the easing curve (edge delays ~2x center)
- So `delta_y=10` → ~200ms total (~20ms avg/tick), and `delta_y=50` → still ~200ms total (~4ms avg/tick)

This keeps large scrolls snappy instead of blocking input for 500ms+.

See 627054d.

PR #184, opened 2026-03-23
Sync upstream - v39/v43 (20260512)

archandatta:
## Summary
- Synced `sync/upstream-release` with new private `origin/main` commits, including the Go wrapper rollout, CapMonster v1.14.13 refresh, and vuln-remediation workflow updates.
- Merged upstream `main`, bringing in `#206 ci: use centralized vuln remediation workflow from infra` and `#234 Rewrite chromium wrapper.sh as a Go binary`.
- Checked `kernel-browser` before the upstream merge; the sync branch was already on the latest release `v145.0.7632.75-r15`, so no browser bump commit was needed.
- Image versions are `headless v39` and `headful v43`.

## Sync details
- Date of sync: `2026-05-12`
- Private sync: merged new commits from `origin/main` into `sync/upstream-release`
- Kernel-browser: already up to date at `v145.0.7632.75-r15`
- Upstream merge: complete
- New image versions: headless `v39`, headful `v43`

## Merge conflicts resolved
- `images/chromium-headful/Dockerfile`: `manual` - kept the private `shared/extensions` copy while retaining upstream image changes.
- `images/chromium-headless/image/Dockerfile`: `manual` - combined upstream Envoy install/bootstrap/init changes with the private extensions copy and kept a single baked-cert step.
- `server/cmd/chromium-launcher/main.go`: `manual` - preserved private CapMonster/kernel extension preparation and forward-proxy support.
- `server/cmd/wrapper/chromium.go`: `manual` - kept private headless stealth flags while letting `chromium-launcher` remain the single source of `--headless=new`.
- `server/cmd/wrapper/main.go`: `manual` - preserved private GPU hooks, WebRTC startup sequencing, and forward-proxy readiness while merging upstream wrapper orchestration changes.
- `server/cmd/wrapper/supervisord.go`: `manual` - resolved the comment-only conflict to match the retained fatal boot-path scripts.
- `socket.yml`: `manual` - kept both ignore paths: `scripts/` and `images/chromium-headful/client/`.

## Validation
- `cd server && PATH=/usr/local/go/bin:$PATH && go test -v -race <all non-e2e packages>`
- The first full run hit a timeout in `github.com/kernel/kernel-images/server/lib/cdpmonitor` (`TestConsoleEvents/non_string_args`); a targeted rerun passed, and a second full non-e2e run passed cleanly.


rgarcia (commented):
nice direction on the pipeline/schema. one thing i'd consider before cementing `BrowserEvent` v1: if this is eventually going to drive both capture controls (for example: "turn on cdp console only" or "turn on network request/response capture at headers-only detail") and subscriptions, it might be worth making those selector dimensions first-class in the envelope instead of encoding too much into a single `type` string.

concretely, i think i'd lean toward:
- keeping the primary event identity semantic, e.g. `console.log`, `network.request`, `input.click`
- adding explicit provenance fields like `source_kind` (`cdp`, `kernel_api`, `extension`, `local_process`) plus `source_name` / `source_event` (for example `Runtime.consoleAPICalled`)
- adding an explicit `detail_level` (`minimal`, `default`, `verbose`, `raw`)
- possibly making `category` first-class too instead of deriving it from the type prefix

i probably would *not* use raw `Runtime.*` / `Network.*` as the primary type, since that makes future non-cdp producers feel awkward/second-class. i think the semantic-type + provenance split ages better if we later want to emit events from things like:
- third-party extensions running in the browser and talking to localhost
- vm-local helper processes/programs running alongside the browser
- server/api-driven tool actions like screenshot/input/recording events

that shape also gives the system a much more natural control surface for both capture config and subscriptions, since selectors can operate directly on stable fields like `category`, `topic`, `source_kind`, and `detail_level` instead of needing to parse overloaded event names.

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,110 @@
...
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+// RingBuffer is a fixed-capacity circular buffer with closed-channel broadcast fan-out.
+// Writers never block regardless of reader count or speed.
+// Readers track their position by seq value (not ring index) and receive an
+// events_dropped synthetic BrowserEvent when they fall behind the oldest retained event.
+type RingBuffer struct {
```
Sayan-, line 10: makes sense to handroll. the combination of non-blocking writes with eviction, multi-reader fan-out with independent positions, synthetic drop notification, and context-aware blocking reads is pretty niche, no standard Go library covers it without significant wrapping.

server/lib/events/filewriter.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+)
+
+// FileWriter is a per-category JSONL appender. It opens each log file lazily on
+// first write (O_APPEND|O_CREATE|O_WRONLY) and serialises concurrent writes
+// within a category with a single mutex.
+type FileWriter struct {
```
Sayan-, line 13: doc comment says "serialises concurrent writes within a category" but the mutex is global across all categories. worth either making the comment accurate or switching to per-category mutexes.

server/lib/events/filewriter.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		fw.files[cat] = f
+	}
+
+	data, err := json.Marshal(ev)
+	if err != nil {
+		return fmt.Errorf("filewriter: marshal: %w", err)
+	}
+
+	var buf bytes.Buffer
```
Sayan-: nit: the `bytes.Buffer` alloc per write is unnecessary since we're under the mutex. can just `f.Write(data)` + `f.Write([]byte{'\n'})` directly.

server/lib/events/events_test.go
```diff
@@ -0,0 +1,474 @@
...
+
+	// Data must be valid JSON with a "dropped" count
+	require.NotNil(t, first.Data)
+	assert.True(t, json.Valid(first.Data))
+	assert.Contains(t, string(first.Data), `"dropped"`)
+}
+
+// TestConcurrentReaders: 3 readers subscribe before publish; publish 5 events;
+// each reader independently reads all 5; no reader affects another.
+func TestConcurrentReaders(t *testing.T) {
```
Sayan-, line 287: would be good to add a test where publishing and reading happen in parallel to exercise the locking/contention paths under `go test -race`.

Sayan- (commented):
focused review on the pipeline since raf had some feedback to the event schema!

server/lib/events/event.go
```diff
@@ -0,0 +1,76 @@
...
+	URL              string          `json:"url,omitempty"`
+	Data             json.RawMessage `json:"data,omitempty"`
+	Truncated        bool            `json:"truncated,omitempty"`
+}
+
+// truncateIfNeeded marshals ev and returns the (possibly truncated) event together
+func truncateIfNeeded(ev BrowserEvent) (BrowserEvent, []byte) {
+	data, err := json.Marshal(ev)
+	if err != nil {
+		return ev, data
```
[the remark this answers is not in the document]
Sayan-: ==

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,98 @@
...
+		notify := r.rb.notify
+		oldest := r.rb.oldestSeq()
+		written := r.rb.written
+
+		if r.nextSeq < oldest {
+			dropped := oldest - r.nextSeq
+			r.nextSeq = oldest
+			r.rb.mu.RUnlock()
+			data := json.RawMessage(fmt.Sprintf(`{"dropped":%d}`, dropped))
+			return BrowserEvent{Type: "events.dropped", Category: CategorySystem, Source: SourceKernelAPI, Data: data}, nil
```
Sayan-: On final review I'm wondering if the drop sentinel should be a separate type from `BrowserEvent`. A drop notification is stream metadata, not browser content.

A deeper dive a la opus:

The current sentinel has fabricated/zero fields (`Seq: 0`, `Ts: 0`, `Category: "system"`, `Source: "kernel_api"`) that aren't actually observed from the browser. Since there are no consumers yet, now's the cheapest time to split this:

```go
type ReadResult struct {
    Event   *BrowserEvent // nil when Dropped > 0
    Dropped uint64        // count of events lost at this point
}
```

This makes the contract compile-time safe consumers can't accidentally serialize a drop as a real event because `Event` is nil. The transport layer (e.g. a future WebSocket handler) can decide how to represent drops on the wire independently.

server/lib/events/capturesession.go
```diff
@@ -0,0 +1,68 @@
...
+	"sync/atomic"
+	"time"
+)
+
+// Pipeline glues a RingBuffer and a FileWriter into a single write path
+type Pipeline struct {
+	mu               sync.Mutex
+	ring             *RingBuffer
+	files            *FileWriter
+	seq              atomic.Uint64
```
Sayan-: might be simpler to just make this an `uint64` since we only mutate it while holding the lock. nbd

server/lib/events/pipeline.go
```diff
@@ -0,0 +1,68 @@
+package events
+
+import (
+	"log/slog"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Pipeline glues a RingBuffer and a FileWriter into a single write path
```
Sayan-: may be helpful to capture the intended use of the pipeline wrt lifecycle (e.g. can you stop and restart the same pipeline, do you need to send a terminal event before closing, etc)

Sayan- (commented):
changes look great!

server/lib/events/pipeline.go
```diff
@@ -0,0 +1,68 @@
...
+func NewPipeline(ring *RingBuffer, files *FileWriter) *Pipeline {
+	p := &Pipeline{ring: ring, files: files}
+	empty := ""
+	p.captureSessionID.Store(&empty)
+	return p
+}
+
+// Start sets the capture session ID that will be stamped on every subsequent
+// published event
+func (p *Pipeline) Start(captureSessionID string) {
```
rgarcia: i think we're still missing a separation between subscription selectors and native producer config. some producers will be native to the image server and controllable by us (like cdp), while others may be external and not controllable at all. because of that, i'm not sure subscriptions should implicitly mean "turn this producer on/reconfigure it". i think the api may want subscriptions to stay a pure consumer-side filter, with native producer enablement/config modeled explicitly and separately.

server/lib/events/event.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// BrowserEvent is the canonical event structure for the browser capture pipeline.
+type BrowserEvent struct {
+	CaptureSessionID string          `json:"capture_session_id"`
+	Seq              uint64          `json:"seq"`
+	Ts               int64           `json:"ts"`
+	Type             string          `json:"type"`
+	Category         EventCategory   `json:"category"`
+	Source           Source          `json:"source"`
```
rgarcia: this is starting to feel like provenance wants to be an object rather than multiple top-level fields. i'd consider making source structured and moving cdp-specific context under source.metadata, e.g.

~~~json
{
  "type": "console.log",
  "category": "console",
  "detail_level": "standard",
  "source": {
    "kind": "cdp",
    "event": "Runtime.consoleAPICalled",
    "metadata": {
      "target_id": "...",
      "cdp_session_id": "...",
      "frame_id": "...",
      "parent_frame_id": "..."
    }
  },
  "data": { ... }
}
~~~

that keeps the top level focused on stable cross-producer fields and makes non-cdp producers feel first-class too.

server/lib/events/event.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	DetailMinimal DetailLevel = "minimal"
+	DetailDefault DetailLevel = "default"
+	DetailVerbose DetailLevel = "verbose"
+	DetailRaw     DetailLevel = "raw"
+)
+
+// BrowserEvent is the canonical event structure for the browser capture pipeline.
+type BrowserEvent struct {
+	CaptureSessionID string          `json:"capture_session_id"`
```
rgarcia: what is capture_session_id intended to identify? this reads more like pipeline/native-producer lifecycle metadata than part of the portable event schema. if resume/subscription are meant to work across heterogeneous producers, i'd be careful about baking a producer-owned session id into every event.

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,98 @@
...
+		return 0
+	}
+	return rb.written - uint64(len(rb.buf))
+}
+
+// NewReader returns a Reader positioned at publish index 0
+// If the ring has already published events, the reader will receive an
+// events_dropped BrowserEvent on the first Read call if it has fallen behind
+// the oldest retained event
+func (rb *RingBuffer) NewReader() *Reader {
```
rgarcia: i think the resumable story wants to show up in the api here eventually. right now NewReader() always starts from 0, so there isn't yet an explicit way to resume from a previously persisted cursor/offset after reconnect.

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,98 @@
...
+	return &Reader{rb: rb, nextSeq: 0}
+}
+
+// Reader tracks an independent read position in a RingBuffer.
+type Reader struct {
+	rb      *RingBuffer
+	nextSeq uint64 // publish index, not BrowserEvent.Seq
```
rgarcia: more fundamentally, i think Event.Seq probably wants to be the public stream offset / resume token. right now the reader tracks a separate publish-index concept (nextSeq / written), which creates two seq-like notions in the model. i'd strongly consider aligning the reader/ring terminology and api around Event.Seq instead of introducing a second cursor concept.

server/lib/events/ringbuffer.go
rgarcia: i'm a little nervous about representing drops as a normal Event. this feels like stream/transport metadata rather than producer-emitted content. maybe Read should return a separate result type, e.g. ReadResult{Event *Event, Dropped uint64}, where Event is nil when the reader has fallen behind. then the transport layer can choose how to surface drops on the wire without baking them into the canonical event schema.

server/lib/events/event.go
```diff
@@ -0,0 +1,76 @@
...
+	SourceKernelAPI    Source = "kernel_api"
+	SourceExtension    Source = "extension"
+	SourceLocalProcess Source = "local_process"
+)
+
+type DetailLevel string
+
+const (
+	DetailMinimal DetailLevel = "minimal"
+	DetailDefault DetailLevel = "default"
```
rgarcia: nit: if this is meant to be a concrete level, i'd consider renaming default to standard. minimal/standard/verbose/raw reads like an actual ladder, whereas default feels more like a policy alias whose meaning could vary by producer.

server/lib/events/event.go
```diff
@@ -0,0 +1,76 @@
...
+	DetailVerbose DetailLevel = "verbose"
+	DetailRaw     DetailLevel = "raw"
+)
+
+// BrowserEvent is the canonical event structure for the browser capture pipeline.
+type BrowserEvent struct {
```
rgarcia: nit: i wonder if Event would be a cleaner name than BrowserEvent. it feels more future-proof if extension/local-process/server-native producers end up as first-class peers.

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,98 @@
...
+
+// RingBuffer is a fixed-capacity circular buffer with closed-channel broadcast fan-out.
+// Writers never block regardless of reader count or speed.
+// Readers track their position by seq value (not ring index) and receive an
+// events_dropped synthetic BrowserEvent when they fall behind the oldest retained event.
+type RingBuffer struct {
+	mu      sync.RWMutex
+	buf     []BrowserEvent
+	head    int    // next write position (mod cap)
+	written uint64 // total ever published (monotonic)
```
rgarcia: nit: could we make this comment a bit more explicit? "written" took me a second to parse. something like "monotonic count of publishes into the logical stream" would make it clearer that this is not bytes-written and not Event.Seq.

server/lib/events/ringbuffer.go
```diff
@@ -0,0 +1,98 @@
...
+	notify  chan struct{}
```
rgarcia: nit: "notify" feels a bit ambiguous here since it doesn't say who is being notified. i'd consider renaming this to something reader-scoped like readerWake or readerNotify, plus a short comment that it's closed-and-replaced on each publish to wake blocked readers.

rgarcia (changes requested):

server/lib/events/pipeline.go
```diff
@@ -0,0 +1,70 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Pipeline is a single-use write path that wraps events in envelopes and fans
+// them out to a FileWriter (durable) and RingBuffer (in-memory). Call Start
+// once with a capture session ID, then Publish concurrently. Close flushes the
+// FileWriter; there is no restart or terminal event.
+type Pipeline struct {
```
rgarcia: one design question: now that this object owns seq allocation, envelope stamping, reader creation, and capture_session_id, i wonder if Pipeline is too generic a name. it feels more like a session-scoped event stream than a reusable pipeline primitive. would something like CaptureSession or CaptureStream be clearer? relatedly, if capture_session_id is meant to identify exactly one capture session, i think i'd prefer passing it into the constructor and dropping Start(), so the type can't exist in a pre-start / empty-session state and can't accidentally be reused across sessions.

server/lib/events/event.go
```diff
@@ -0,0 +1,91 @@
...
+	DetailMinimal  DetailLevel = "minimal"
+	DetailStandard DetailLevel = "standard"
+	DetailVerbose  DetailLevel = "verbose"
+	DetailRaw      DetailLevel = "raw"
+)
+
+// Event is the portable event schema. It contains only producer-emitted content;
+// pipeline metadata (seq, capture session) lives on the Envelope.
+type Event struct {
+	Ts          int64           `json:"ts"`
```
rgarcia: nit: could we document `ts` explicitly? right now it isn't obvious whether this is unix seconds, millis, or nanos. i do wonder if millisecond resolution is a bit coarse for this stream since multiple events can plausibly share the same ms; if we want higher fidelity, maybe consider unix micros.

rgarcia (approved):
🔥

server/lib/events/pipeline.go
archandatta: went with `CaptureSession`!

rgarcia (approved):

PR #194, opened 2026-04-01
[kernel 1116] browser logging: cdp pipeline

archandatta:
Scaffolding of the CDP Pipeline implementation on the functions are in -> https://github.com/kernel/kernel-images/pull/202 

---

> [!NOTE]
> **Medium Risk**
> Adds new capture-session lifecycle endpoints and introduces a new in-process event pipeline with file/ring-buffer writers, plus new shutdown/constructor wiring; concurrency and lifecycle handling are the main risk areas. CDP monitor is currently scaffolded, but the new API surface and persistence paths still warrant careful review.
> 
> **Overview**
> Adds an **event capture session** feature, exposing `GET/POST/PATCH/DELETE /events/capture_session` to start/stop and reconfigure capture (including category filtering) and report session `id`, `status`, `seq`, and `created_at`.
> 
> Wires a new `events.CaptureSession` and `cdpmonitor.Monitor` into `ApiService`/`main`, including lifecycle cancellation and shutdown cleanup, and refactors the events pipeline to be reusable (`Start`/`Stop`), category-filtered, and backed by a validated/resettable ring buffer plus a filename-based JSONL writer.
> 
> Regenerates OpenAPI (`oapi.go`, `openapi.yaml`) and updates/extends tests to cover the new capture session handlers and the updated events/ring buffer/file writer behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33c07d36ee6a11cc1a69515dda8f24902322ba79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/cmd/api/api/api.go
```diff
@@ -68,11 +70,24 @@ type ApiService struct {
 	// xvfbResizeMu serializes background Xvfb restarts to prevent races
 	// when multiple CDP fast-path resizes fire in quick succession.
 	xvfbResizeMu sync.Mutex
+
+	// CDP event pipeline and cdpMonitor.
+	eventsPipeline *events.Pipeline
```
archandatta: Will change this to `CaptureSession` once https://github.com/kernel/kernel-images/pull/184 is merged

server/cmd/api/api/api.go
```diff
@@ -313,5 +334,9 @@ func (s *ApiService) ListRecorders(ctx context.Context, _ oapi.ListRecordersRequ
 }
 
 func (s *ApiService) Shutdown(ctx context.Context) error {
+	s.monitorMu.Lock()
```
hiroTamada, line 342: Is this going to add a latency in delete browsers request? If so, how much?

server/cmd/api/main.go
```diff
@@ -120,6 +128,10 @@ func main() {
 		w.Header().Set("Content-Type", "application/json")
 		w.Write(jsonData)
 	})
+	// capture events
+	r.Post("/events/start", apiService.StartCapture)
```
hiroTamada: Do we have openapi.yaml spec for this? I dont see it

server/cmd/api/api/events.go
```diff
@@ -0,0 +1,35 @@
...
+// Generates a new capture session ID, seeds the pipeline, then starts the
+// CDP monitor. If already running, the monitor is stopped and
+// restarted with a fresh session ID
+func (s *ApiService) StartCapture(w http.ResponseWriter, r *http.Request) {
+	s.monitorMu.Lock()
+	defer s.monitorMu.Unlock()
+
+	s.captureSession.Start(uuid.New().String())
+
+	if err := s.cdpMonitor.Start(context.Background()); err != nil {
```
rgarcia: the comment says this restarts an existing monitor, but the handler never checks/stops an already-running one. might be worth making that behavior explicit here instead of relying on `Start()` side effects.

server/cmd/api/api/events.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/google/uuid"
+	"github.com/onkernel/kernel-images/server/lib/logger"
+)
+
+// StartCapture handles POST /events/start.
+// Generates a new capture session ID, seeds the pipeline, then starts the
+// CDP monitor. If already running, the monitor is stopped and
+// restarted with a fresh session ID
+func (s *ApiService) StartCapture(w http.ResponseWriter, r *http.Request) {
```
rgarcia: question: is a no-arg `/events/start` intentional? i'd expect callers to want some capture config here over time (categories, detail level, screenshot/network verbosity, etc.), or at least an explicit note that this is a fixed server-defined profile for now.

server/cmd/api/main.go
rgarcia: agree with Hiro, these should be in the openapi spec, and we should think carefully about how users might want to configure the CDP event capture in that API.

server/cmd/api/main.go
```diff
@@ -90,12 +91,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Construct events pipeline
+	eventsRing := events.NewRingBuffer(1024)
+	eventsFileWriter := events.NewFileWriter("/var/log")
```
rgarcia: can we write these under `/var/log/kernel/` instead of directly under top-level `/var/log`? otherwise we'll pollute the shared log root. if we do that, `NewFileWriter` probably wants to ensure the directory exists too.

server/lib/cdpmonitor/monitor.go
```diff
@@ -0,0 +1,41 @@
...
+type PublishFunc func(ev events.Event)
+
+// Monitor manages a CDP WebSocket connection with auto-attach session fan-out.
+// Single-use per capture session: call Start to begin, Stop to tear down.
+type Monitor struct {
+	running atomic.Bool
+}
+
+// New creates a Monitor. displayNum is the X display for ffmpeg screenshots.
+func New(_ UpstreamProvider, _ PublishFunc, _ int) *Monitor {
```
rgarcia, line 26: i don't think we'll need `displayNum` for a pure CDP monitor, so i'd remove this until actual implementation lands.

server/lib/events/capturesession.go
```diff
@@ -3,22 +3,32 @@ package events
...
+// fans them out to a FileWriter (durable) and RingBuffer (in-memory). Call Start
+// once with a capture session ID, then Publish concurrently. Close flushes the
+// FileWriter; there is no restart or terminal event.
 type CaptureSession struct {
 	mu               sync.Mutex
 	ring             *RingBuffer
 	files            *FileWriter
-	seq              uint64
-	captureSessionID string
+	seq              atomic.Uint64
```
rgarcia: this seems to be reverting a couple decisions from the earlier events work: switching timestamps back to milliseconds, and switching `seq` / `captureSessionID` back to atomics. was that intentional? i thought we had landed on microsecond timestamps and plain fields here.

server/lib/events/capturesession.go
```diff
@@ -3,22 +3,32 @@ package events
...
+	captureSessionID atomic.Pointer[string]
 }
 
-func NewCaptureSession(captureSessionID string, ring *RingBuffer, files *FileWriter) *CaptureSession {
-	return &CaptureSession{ring: ring, files: files, captureSessionID: captureSessionID}
+func NewCaptureSession(ring *RingBuffer, files *FileWriter) *CaptureSession {
```
rgarcia: `CaptureSession` still feels like it's exposing a lot of implementation detail through the constructor. i can see passing a directory/config string in, but the ring buffer and file writer feel like internals that `CaptureSession` should probably own unless we expect callers to swap implementations independently.

server/lib/events/capturesession.go
```diff
@@ -3,22 +3,32 @@ package events
...
+	s := &CaptureSession{ring: ring, files: files}
+	empty := ""
+	s.captureSessionID.Store(&empty)
+	return s
+}
+
+// Start sets the capture session ID stamped on every subsequent envelope.
+func (s *CaptureSession) Start(captureSessionID string) {
```
rgarcia: i'd add a guard against starting an already-started pipeline, since we're saying that `Start` can only be called once.

server/lib/events/event.go
```diff
@@ -50,7 +51,7 @@ const (
 // Event is the portable event schema. It contains only producer-emitted content;
 // pipeline metadata (seq, capture session) lives on the Envelope.
 type Event struct {
-	Ts          int64           `json:"ts"` // Unix microseconds (µs since epoch)
+	Ts          int64           `json:"ts"`
```
rgarcia: this comment is useful imo.

server/lib/events/event.go
```diff
@@ -67,6 +68,30 @@ type Envelope struct {
 	Event            Event  `json:"event"`
 }
 
+// CategoryFor derives an EventCategory from an event type string.
+// It splits on the first dot and maps the prefix to a category.
+func CategoryFor(eventType string) EventCategory {
```
rgarcia: if this helper is only meant for the CDP producer, i'd keep it local to that producer rather than adding it to the shared `events` package. putting it here makes it look like `type -> category` is a general event-model rule, and i don't think we want to tightly couple those fields like that.

rgarcia (changes requested):

server/cmd/api/api/events.go
archandatta: Ah yeah this was just a scaffolding I did for the actual implementation. I should've mentioned that there is a chained PR https://github.com/kernel/kernel-images/pull/202 that is capturing this https://github.com/kernel/kernel-images/pull/202/changes#diff-0ddb00e75e5f21a1deb16fb852c7f352f234bbf0b9ddcc3d11639b1d3293c572R90

server/lib/events/capturesession.go
archandatta: Refer to earlier comment https://github.com/kernel/kernel-images/pull/194#discussion_r3046823739

server/lib/events/capturesession.go
archandatta: Ah I think this was a bad merge

server/lib/cdpmonitor/monitor.go
archandatta: this is handled in https://github.com/kernel/kernel-images/pull/202

server/cmd/api/api/api.go
archandatta: Shouldn't add any latency, shutdown holds the lock briefly to stop the CDP monitor and close the capture session i think both are fast synchronous teardown calls

server/cmd/api/api/events.go
archandatta: yeah good point, added a body field to be able to send something like
```
  {
    "categories": ["console", "network"],
    "detailLevel": "verbose"
  }
```

server/lib/events/capturesession.go
archandatta: good point - add the ring buffer and file writer to be internal to CaptureSession

server/openapi.yaml
```diff
@@ -1194,8 +1194,71 @@ paths:
           $ref: "#/components/responses/BadRequestError"
         "500":
           $ref: "#/components/responses/InternalError"
+  /events/start:
```
rgarcia: i think the right framing here is that this request is configuring our single type of "native" producer, not exposing CDP itself as a first-class public producer. today that native producer mostly happens to do things over CDP, but over time i could see it also emitting events from screenshots, local processes, or VM/system monitoring, and i think that still makes sense to model as one native capture source with one config shape.

what feels off to me is the action-style API shape of `/events/start` and `/events/stop`. i think we should model capture sessions as resources instead, so the API captures the lifecycle of a native capture session more directly: create one, inspect it, update it mid-session, stop it, and eventually stream from it. that gives us a cleaner place for the config we want now, and it leaves room for where this may go in the future as the native producer expands beyond just CDP.

one other reason i think this resource model is cleaner is that we may eventually want external producers to publish into the pipeline too, e.g. an extension posting events into an API endpoint. if we do that, i could imagine something like `POST /events` for externally-produced events, while capture-session lifecycle remains its own thing. that separation feels cleaner than baking everything into `/events/start`.

something like:

```yaml
/events/capture_sessions:
  post:
    operationId: createCaptureSession

/events/capture_sessions/{capture_session_id}:
  get:
    operationId: getCaptureSession
  patch:
    operationId: updateCaptureSession
  delete:
    operationId: deleteCaptureSession

/events/capture_sessions/{capture_session_id}/stream:
  get:
    operationId: streamCaptureSession
```

then `CreateCaptureSessionRequest` / `UpdateCaptureSessionRequest` can just carry the native-producer knobs (`categories`, `detailLevel`, etc.), while externally-produced events can be modeled separately.

rgarcia (changes requested):

server/openapi.yaml
archandatta: Yep makes sense! The native-producer model def maps more cleanly to what we already have internally since CaptureSession already has an ID, config, seq counter, and a NewReader() for streaming. The current /events/start and /events/stop just hides that right now. 

One thought is when `POST /events/capture_sessions` is called if a session already exists, we'd just return a 409 and requiring an explicit DELETE first, rather than silently replacing. imo makes the lifecycle explicit and avoids surprising a caller who didn't realize a session was already running. I'll also remove /events/start and /events/stop.

archandatta: Changes made here -> [1cddf53](https://github.com/kernel/kernel-images/pull/194/commits/1cddf5377ae7a188aaf264501890f80fe7cea5a4)

server/lib/events/filewriter.go
```diff
@@ -16,9 +16,12 @@ type FileWriter struct {
 	dir   string
 }
 
-// NewFileWriter returns a FileWriter that writes to dir
-func NewFileWriter(dir string) *FileWriter {
-	return &FileWriter{dir: dir, files: make(map[EventCategory]*os.File)}
+// NewFileWriter returns a FileWriter that writes to dir, creating it if needed.
+func NewFileWriter(dir string) (*FileWriter, error) {
```
rgarcia: nit: if this is only used within events as an implementation detail, then we don't need to export any of the FileWriter stuff

server/lib/events/filewriter.go
```diff
@@ -42,9 +45,12 @@ func (fw *FileWriter) Write(env Envelope, data []byte) error {
 		fw.files[cat] = f
 	}
 
-	if _, err := f.Write(append(data, '\n')); err != nil {
+	if _, err := f.Write(data); err != nil {
 		return fmt.Errorf("filewriter: write: %w", err)
 	}
+	if _, err := f.Write([]byte{'\n'}); err != nil {
```
rgarcia, line 51: it looks like all Write() cares about is the category of the event in order to write a filename. I'd tighten the interface from `Write(env Envelope, data []byte) error` to `Write(filename string, data []byte) error`

and then it's more generic / ess coupled to the large surface area of the Envelope and Event types

server/lib/events/ringbuffer.go
```diff
@@ -16,13 +16,31 @@ type RingBuffer struct {
 }
 
 func NewRingBuffer(capacity int) *RingBuffer {
+	if capacity <= 0 {
+		panic("events: ring buffer capacity must be > 0")
```
rgarcia: 🙅 panic() please, return an error

server/openapi.yaml
```diff
@@ -1194,8 +1194,183 @@ paths:
           $ref: "#/components/responses/BadRequestError"
         "500":
           $ref: "#/components/responses/InternalError"
+  /events/capture_sessions:
```
rgarcia: sorry for going back and forth on this--but I think a singleton capture session is probably the long-term model here, so the API should probably say that directly instead of pretending this is a collection resource. something like `GET /events/capture_session` for status, `POST /events/capture_session` to start, `PATCH /events/capture_session` to update config, and `DELETE /events/capture_session` to stop would read a lot more clearly. right now the plural `/events/capture_sessions/{id}` shape suggests true multi-session support even though the implementation intentionally enforces a single global session, which I think is correct. Multiple capture sessions will get confusing if we also want the event stream to be publishable to from external producers

server/openapi.yaml
```diff
@@ -1194,8 +1194,183 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/CaptureSession"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+    patch:
+      summary: Update a capture session
+      description: >
+        Updates the configuration of an active capture session. The monitor
+        continues running; filtering is applied at publish time.
```
rgarcia: somewhat confusing description--I'd just leave it as "Updates the active capture session" since `monitor` and `filtering` have unclear meanings from an API consumer's point of view

server/openapi.yaml
```diff
@@ -1194,8 +1194,183 @@ paths:
...
+              - page
+              - interaction
+              - liveview
+              - captcha
+              - system
+        detail_level:
+          type: string
+          description: >
+            Default detail level for captured events. Individual events
+            may override this based on their content.
```
rgarcia: i think detail_level right now is somewhat ambiguous and we can kill it. keep categories as the broad top-level filter, and then in the future we can add nested per-category knobs that will be much easier to understand and document, e.g.

```
network.capture_response_bodies
interaction.capture_dom_instrumentation
interaction.capture_user_actions
```

server/lib/events/event.go
```diff
@@ -21,6 +21,33 @@ const (
...
+		m[c] = struct{}{}
+	}
+	return m
+}()
+
+// ValidCategory reports whether c is a known EventCategory.
+func ValidCategory(c EventCategory) bool {
+	_, ok := validCategories[c]
+	return ok
+}
```
rgarcia, line 43: just noticed Event as a URL field which I think can be omitted and included in the Data field when relevant

URL isn't relevant for computer actions / API events

rgarcia (changes requested):
api binary needs to be removed

server/lib/events/ringbuffer.go
archandatta: Sorry should've caught this! -> [a8bdeaf](https://github.com/kernel/kernel-images/pull/194/commits/a8bdeafea38f7387abe3e464e60f3ad04a13b81d)

server/lib/events/filewriter.go
archandatta: [c6dd362](https://github.com/kernel/kernel-images/pull/194/commits/c6dd36233230aa60041803608b3119c5d0a5a27c)

server/lib/events/filewriter.go
archandatta: [ff8bddf](https://github.com/kernel/kernel-images/pull/194/commits/ff8bddf0a5aa044b19e8c8e6ecc4fbec6741d548)

server/openapi.yaml
archandatta: [8f88ed0](https://github.com/kernel/kernel-images/pull/194/commits/8f88ed0a056cfb0a6dd3ef3d61a8a072d1cdca19)

server/openapi.yaml
archandatta: [8f88ed0](https://github.com/kernel/kernel-images/pull/194/commits/8f88ed0a056cfb0a6dd3ef3d61a8a072d1cdca19)

server/openapi.yaml
archandatta: [8f88ed0](https://github.com/kernel/kernel-images/pull/194/commits/8f88ed0a056cfb0a6dd3ef3d61a8a072d1cdca19)

server/cmd/api/api/capture_session.go
```diff
@@ -0,0 +1,149 @@
...
+	if s.captureSession.ID() != "" {
+		return oapi.StartCaptureSession409JSONResponse{ConflictErrorJSONResponse: oapi.ConflictErrorJSONResponse{Message: "a capture session is already active"}}, nil
+	}
+
+	cfg, err := captureConfigFrom(req.Body)
+	if err != nil {
+		return oapi.StartCaptureSession400JSONResponse{BadRequestErrorJSONResponse: oapi.BadRequestErrorJSONResponse{Message: err.Error()}}, nil
+	}
+
+	id := uuid.New().String()
```
rgarcia: let's use github.com/nrednav/cuid2 to be consistent with the rest of the api (and remove format: uuid for this id field in the openapi.yaml spec)

rgarcia (approved):
one small thing but otherwise i think this is good to go!

server/cmd/api/main.go
```diff
@@ -90,12 +91,24 @@ func main() {
...
+	}
+
 	apiService, err := api.New(
 		recorder.NewFFmpegManager(),
 		recorder.NewFFmpegRecorderFactory(config.PathToFFmpeg, defaultParams, stz),
 		upstreamMgr,
 		stz,
 		nekoAuthClient,
+		captureSession,
+		config.DisplayNum,
```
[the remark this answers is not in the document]
Sayan-: ==

Sayan- (approved):
I think the last bugbot comment is worth resolving. Thanks for iterating - interfaces, components, and integration are looking great!

Sayan- (approved):

PR #213, opened 2026-04-13
[kernel-1116] Add CDP Monitor

archandatta:
Introduces the foundational layer of the CDP monitor as a standalone reviewablechunk. No Monitor struct wiring, just the primitives that everything else builds on.

- types.go: CDP wire format (cdpMessage), all event type constants, internal state structs (networkReqState, targetInfo, CDP param shapes).

- util.go: Console arg extraction, MIME allow-list (isCapturedMIME), resource type filter (isTextualResource), per-MIME body size caps (bodyCapFor), UTF-8-safe body truncation (truncateBody).

- computed.go: State machine for the three derived events: network_idle (500ms debounce after all requests finish), layout_settled (1s after page_load with no layout shifts), navigation_settled (fires once all three flags converge). Timer invalidation via navSeq prevents stale AfterFunc callbacks from publishing for a previous navigation.

- domains.go: isPageLikeTarget predicate (pages and iframes get Page.* / PerformanceTimeline.*; workers don't), bindingName constant, interaction.js embed.

- interaction.js: Injected script tracking clicks, keydowns, and scroll-settled events via the __kernelEvent CDP binding.

---

> [!NOTE]
> **High Risk**
> Large new CDP capture subsystem that streams network/console data (including headers/bodies) and manages reconnect/timers across goroutines, which is complex and can impact stability, performance, and sensitive-data exposure if misconfigured.
> 
> **Overview**
> Adds a full `cdpmonitor` implementation that connects to Chrome’s DevTools WebSocket, auto-attaches to targets, translates CDP notifications into typed `events.Event`s (console, network, page, interaction), and publishes monitor lifecycle events (disconnect/reconnect/init-failed) with capped-exponential reconnect backoff.
> 
> Introduces per-session computed state machines (`network_idle`, `page_layout_settled`, `page_navigation_settled`), interaction tracking via embedded `interaction.js` with rate-limiting and sensitive-input suppression, and screenshot capture via `ffmpeg` with size downscaling and 2s rate limiting.
> 
> Updates API wiring to pass a `slog` logger into `cdpmonitor.New`, abstracts `ApiService.cdpMonitor` behind an interface for easier stubbing in tests, and adds extensive fixtures/tests to validate CDP wire-type roundtrips and monitor behavior (events, redirects, detach handling, TTL sweeps, and reconnects).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cc19d8cfa776bf620b72008abad2550aa8b8666. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hiroTamada (dismissed):
solid foundation overall — the computed event state machine and timer invalidation logic are well designed. one blocking concern around sensitive data capture in interaction.js.

server/lib/cdpmonitor/interaction.js
```diff
@@ -0,0 +1,64 @@
...
+    var t = e.target || {};
+    send(JSON.stringify({
+      type: 'interaction_click',
+      x: e.clientX, y: e.clientY,
+      selector: sel(t), tag: t.tagName || '',
+      text: (t.textContent || '').slice(0, 100)
+    }));
+  }, true);
+
+  document.addEventListener('keydown', function(e) {
```
hiroTamada, line 81: the keydown listener captures every keystroke with no filtering — this is effectively a keylogger. since this gets injected into all browser VMs (including sessions where users type passwords via managed auth), sensitive input like passwords, credit card numbers, and SSNs will flow through the event pipeline and get persisted. at minimum, keystrokes in `<input type="password">` fields should be redacted or skipped. worth also considering fields with `autocomplete="cc-number"`, `name="ssn"`, etc. the click listener's `textContent` capture has a similar (lesser) exposure. this has PCI/SOC2/GDPR implications if the data hits storage.

server/lib/cdpmonitor/types.go
```diff
@@ -0,0 +1,176 @@
...
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// mainSessionUnset is the sentinel stored in mainSessionID before any
+// top-level frameNavigated has been received. Using a sentinel prevents the
+// empty-string zero value from accidentally matching CDP messages that arrive
+// on the root session (sessionID="") before navigation has been recorded.
+const mainSessionUnset = "\x00unset"
```
rgarcia, line 13: this is hard to comment on in isolation while the constant is still unused. i'm questioning whether splitting the PRs at this layer is actually helping in this particular case, since pieces like this sentinel only really make sense once the monitor state machine / routing logic using them is in the diff too.

server/lib/cdpmonitor/types.go
```diff
@@ -0,0 +1,176 @@
...
+
+// Event type constants for all events published by the cdpmonitor.
```
rgarcia: i think this constant block would be a lot easier to reason about with some cosmetic structure: group the names into direct-ish CDP-derived monitor events, computed/synthetic events from `computed.go`, injected page-side interaction events, and monitor lifecycle/internal events, and add short godoc comments for each group (or for the less obvious constants). right now they're all in one flat list, which makes it hard to tell which names are grounded in protocol counterparts vs introduced locally by the monitor.

server/lib/cdpmonitor/types.go
```diff
@@ -0,0 +1,176 @@
...
+	postData     string
+	resourceType string
+	status       int
+	statusText   string
+	resHeaders   json.RawMessage
+	mimeType     string
+	addedAt      time.Time // for TTL eviction
+}
+
+// cdpConsoleArg is a single Runtime.consoleAPICalled argument.
```
rgarcia: i'd like for this first decode layer to be much closer to the CDP protocol / PDL types, and then have a second translation step into monitor-specific event shapes. right now things like `cdpConsoleArg` / `cdpConsoleParams` are custom projections of `Runtime.RemoteObject` / `Runtime.consoleAPICalled`, which makes it harder to audit against the source of truth and easier to accidentally drop fields we may want later. i think mirroring the protocol more directly at the boundary would make this foundation easier to reason about, with the domain-specific simplification happening after decode. the upstream source of truth here is the CDP spec / PDL, e.g. [js_protocol.pdl](https://github.com/ChromeDevTools/devtools-protocol/blob/master/pdl/js_protocol.pdl).

rgarcia (changes requested):
can you combine all the prs into one? I'm finding it hard to review this since most of the code is unused

server/lib/cdpmonitor/types.go
archandatta: Merged the PRs into this one

server/lib/cdpmonitor/types.go
archandatta: Submitted a plan for the changes -> https://github.com/kernel/kernel-images/pull/216

archandatta: addressed -> [0605227](https://github.com/kernel/kernel-images/pull/213/commits/06052274c4d7bd2d0ccbf83cdd6b2d2118ecea24)

server/lib/cdpmonitor/interaction.js
archandatta: addressed -> [a62c403](https://github.com/kernel/kernel-images/pull/213/commits/a62c403a5d2d22545f811351c503cb938b9a393a)

server/lib/cdpmonitor/types.go
archandatta: [a62c403](https://github.com/kernel/kernel-images/pull/213/commits/a62c403a5d2d22545f811351c503cb938b9a393a)

server/lib/cdpmonitor/cdp_proto.go
```diff
@@ -0,0 +1,279 @@
+package cdpmonitor
+
+// Layer-1 PDL-faithful CDP types — one struct per dispatched event, retaining
```
Sayan-, line 3: https://github.com/kernel/kernel-images/pull/216/files#diff-324e08c20aa8c7007a1b64e5c1c1ec620cf5b91eb4eb59ded9a90f9b68510eebR99 mentions version pinning. did we still need that?

Sayan- (dismissed):
Overall looks good! Reading through https://github.com/kernel/kernel-images/pull/216 was super helpful to ground the motivation of some of these changes.

I do think the pending bot comments should be addressed.

From a scalability perspective a few properties I noticed:

- Publish is synchronous and serialized. Every handler blocks on `CaptureSession.Publish`, which holds a session-wide mutex and writes to disk inline.
- Per-response goroutine fanout for body fetches. `handleResponseReceived` spawns one unbounded goroutine per textual response to call `Network.getResponseBody`.
- Screenshots are inline base64 PNGs carried through the same publish path.

I think the units have reasonable coverage. As we get closer to shipping, I'd love to see a consistent stress test against a real page so we can understand latency, memory, and perf implications.

Sayan- (approved):

server/lib/cdpmonitor/monitor.go
```diff
@@ -16,26 +22,571 @@ type UpstreamProvider interface {
...
+func (m *Monitor) getLifecycleCtx() context.Context {
+	m.lifeMu.Lock()
+	ctx := m.lifecycleCtx
+	m.lifeMu.Unlock()
+	return ctx
+}
+
 // Start begins CDP capture. Restarts if already running.
+// Not concurrency-safe; callers must serialize Start calls.
 func (m *Monitor) Start(_ context.Context) error {
```
rgarcia: why is context ignored? `main()` already has a process-scoped signal context, but `api.New()` and `cdpmonitor.Start()` both root their lifecycle in `context.Background()`. Since the monitor is long-lived it shouldn’t inherit an HTTP request context, but should it inherit the application/service context from `main()` instead? Shutdown works today via explicit `apiService.Shutdown(...)`, but threading a parent lifecycle context through would make the ownership model clearer and reduce reliance on manual teardown.

server/lib/cdpmonitor/monitor.go
```diff
@@ -16,26 +22,571 @@ type UpstreamProvider interface {
...
+
 // Monitor manages a CDP WebSocket connection with auto-attach session fan-out.
-// Single-use per capture session: call Start to begin, Stop to tear down.
+//
+// Lock ordering (outer → inner):
+//
+//	restartMu → lifeMu → pendReqMu → computed.mu → pendMu → sessionsMu
+//
+// Never acquire a lock that appears later in this order while holding an
+// earlier one, to prevent deadlock.
```
rgarcia: this is missing a lot of introductory context about what these locks control / what the Monitor struct is responsible for. This file is the entrypoint to the package and reading this I am very lost

maybe a lib/cdpmonitor/README.md would be helpful

server/lib/cdpmonitor/monitor.go
```diff
@@ -16,26 +22,571 @@ type UpstreamProvider interface {
...
 }
 
 // IsRunning reports whether the monitor is actively capturing.
 func (m *Monitor) IsRunning() bool {
 	return m.running.Load()
 }
 
+// getLifecycleCtx returns the current lifecycle context under lifeMu.
+func (m *Monitor) getLifecycleCtx() context.Context {
+	m.lifeMu.Lock()
```
rgarcia: getLifecycleCtx() looks unusual at first glance, but I think it exists to avoid a race with `Start()`/`Stop()` replacing `m.lifecycleCtx` under `lifeMu`. That said, the need for a getter here suggests the lifecycle context is being accessed a bit too ambiently across handlers. Could we instead capture the lifecycle context at the boundary of the async work and pass it down explicitly? That would be easier to reason about than having leaf handlers reach back into Monitor for the current context.

server/lib/cdpmonitor/types.go
```diff
@@ -0,0 +1,117 @@
...
+// process, not from Chrome's network or page domains.
+const (
+	EventInteractionClick = "interaction_click"  // document click (target selector, coords, text)
+	EventInteractionKey   = "interaction_key"    // keydown (key name, target selector)
+	EventScrollSettled    = "scroll_settled"     // 300 ms after the last scroll event on a target
+)
+
+// Monitor lifecycle and internal events — emitted by the monitor itself, not by Chrome.
+const (
+	EventScreenshot             = "screenshot"              // periodic ffmpeg frame capture
```
rgarcia: isn't it event-triggered, not periodic?

server/lib/cdpmonitor/monitor.go
archandatta: Yep that makes sense. passing the context now with the http request context

server/lib/cdpmonitor/monitor.go
archandatta: moved `lifeMu` to the boundary of `dispatchEvent` so that it is passed more explicitly to the handlers that use it

server/lib/cdpmonitor/monitor.go
archandatta: lmk if this is clear enough, or if there other details that would be worth adding! [fd4d4d3](https://github.com/kernel/kernel-images/pull/213/commits/fd4d4d316557b612472c50d122b05059c15d6ab4)

rgarcia:
I did a manual pass against this branch with the review harness, doing a single navigation to `kernel.sh`. Event log from that run: https://gist.github.com/rgarcia/cc2c9bb92b536d7f6659985868a3c56b

A few observations/suggestions from that run:

The event stream needs richer CDP identity metadata before this is reviewable as a browser logging foundation. The monitor already receives enough data to tie network events to a target/frame/navigation, but the published event drops most of it. `Network.requestWillBeSent` includes `requestId`, `loaderId`, `documentURL`, and `frameId`, and the monitor also has `sessionID -> targetInfo` with `targetId`/`targetType`. Today we only publish `cdp_session_id`, method/url/headers/etc. This makes redirects look like duplicate requests and makes it hard to group requests by tab/frame/navigation.

Could we include at least:

- `request_id`
- `loader_id`
- `frame_id`
- `document_url`
- `target_id`
- `target_type`
- redirect metadata, or a clear marker when `requestWillBeSent` represents a redirect continuation

Relatedly, the computed events (`network_idle`, `layout_settled`, `navigation_settled`) currently publish no session/frame/target metadata, and the state machine appears global to the monitor. That makes these events impossible to attribute to a tab from the event stream, and multiple tabs/navigations could interfere with each other. I’d expect computed state to be scoped per page target/session, or at least to carry the current top-level navigation context initialized from `Page.frameNavigated` (`session_id`, `target_id`, `frame_id`, `loader_id`, URL, nav sequence) and publish that context on all synthetic events.

Without this, consumers can’t reliably answer “which tab/page/navigation produced this event?”, which seems core to the usefulness of the CDP monitor.

One note from the log: the first three document `network_request` events are a redirect chain (`http://kernel.sh/` -> `https://kernel.sh/` -> `https://www.kernel.sh/`), not exact duplicates. Including `request_id` and redirect metadata would make that much clearer in the event stream.

archandatta:
@rgarcia feedback addressed! the fields like `layout_settled` and `navigation_settled` are connected with loader id:
```
layout_settled.data.loader_id
      == network_request.data.loader_id (all requests from that navigation)
      == network_response.data.loader_id
```

rgarcia:
I ran another manual pass with the review harness, this time navigating to Hacker News, then opening a new tab and navigating to kernel.sh. Event data: https://gist.github.com/rgarcia/0a60c17ece22e9394bdd9118475a0f5c

Observations:
1. there does not seem to be a consistent "raw CDP passthrough" envelope inside `event.data`.
  `data.event` shows up for `layout_shift` because `PerformanceTimeline.timelineEventAdded` CDP params are shaped as `{ event: ... }`, not because Kernel has a convention that raw CDP lives under `event`. Other events project/rename fields directly into `data`, e.g. `network_request.data.request_id`, `loader_id`, `frame_id`, etc.
  That makes casing hard to reason about and looks sloppy / inconsistent. Right now consumers see a mix of:
    - raw CDP camelCase fields like `frameId`, `nodeId`, `layoutShiftDetails`
    - Kernel-projected snake_case fields like `frame_id`, `resource_type`, `request_id`
    - raw-ish CDP timestamps in some page events
    - synthetic Kernel fields in computed events

    I think we should probably choose one clear convention of always projecting event data into a Kernel-owned snake_case schema to keep the event data clean and consistent


2. Relatedly, I think `server/lib/cdpmonitor/README.md` should grow a consumer-facing schema section. Right now it has a useful taxonomy/internals overview, but not enough for someone consuming the stream to understand fields like `request_id`, `loader_id`, `frame_id`, `target_id`, `cdp_session_id`, `initiator_type`, or the timestamp fields. A short CDP data model primer would help a lot

3. On timestamps: top-level `event.ts` is Unix microseconds, but `dom_content_loaded.data.timestamp` and `page_load.data.timestamp` are CDP monotonic timestamps in seconds. That is useful CDP data, but it is very easy to confuse with the event timestamp. Maybe rename/project it as something like `cdp_timestamp`?

4. `dom_content_loaded` and `page_load` still only have the CDP timestamp in `data`. They do get source metadata for the CDP session/target, but unlike `navigation`, `network_idle`, `layout_settled`, and `navigation_settled`, they do not carry `loader_id`, `frame_id`, URL, or `nav_seq` in the event payload. Since the monitor now tracks nav context, can we stamp that same context onto these page lifecycle events too?


5. `screenshot` still emits as `local_process` with only the PNG payload. Since screenshots are triggered by `Page.loadEventFired` / exceptions, it would be very useful for them to include context like `trigger_event`, `session_id`, `target_id`, `frame_id`, `loader_id`, URL, and `nav_seq`. Otherwise screenshots are hard to correlate with the page/navigation that caused them.

6. No event was fired when I opened a new tab. Not sure what CDP exposes here but I think we should produce something here. Subsequent events were fired once I navigated, which is good.

Overall the new loader/frame/request/target additions are a big improvement. I think the remaining work is mostly making the schema consistent and documented enough for downstream consumers.

rgarcia:
Couple of other things:

1.   These should likely carry current navigation context (session_id, target_id, target_type, frame_id, loader_id, URL, nav_seq) in data, not just metadata:
  • interaction_click
  • interaction_key
  • scroll_settled
  • console_log
  • console_error
  • dom_content_loaded
  • page_load
  • layout_shift
  • screenshot
  Partially addressed:
  • network_request, network_response, network_loading_failed: now mostly good.
  • network_idle, layout_settled, navigation_settled: now good, they carry nav context.
  • navigation: has frame_id, loader_id, URL, but not target_id / target_type in data; those are only in metadata.


2. Event naming is inconsistent. Current event types:
  • console_log, console_error
  • network_request, network_response, network_loading_failed
  • navigation, dom_content_loaded, page_load, layout_shift
  • network_idle, layout_settled, navigation_settled
  • interaction_click, interaction_key, scroll_settled
  • screenshot, monitor_*
  The category already says network, page, interaction, etc., so prefixes like network_ and interaction_ are somewhat redundant. But if event type is intended to be globally unique without category, then prefixes make sense. The issue is inconsistency:
  • interaction_click / interaction_key have interaction_, but scroll_settled does not.
  • network_idle has network_, but layout_settled and navigation_settled do not have page_.
  • page_load has page_, but navigation, layout_shift, layout_settled, navigation_settled do not.
  I think we should choose a convention here. Since these are logs/events that people will grep/query outside typed clients, I’d lean toward globally namespaced names, e.g. `network_request`, `page_navigation`, `interaction_click`, `interaction_scroll_settled`

rgarcia:
new test run where I navigated to hacker news, then opened a new tab and did some kernel.sh things, then went back to the HN tab and clicked into an openai blog post:  https://gist.github.com/rgarcia/a0c7670d3c3850a7fdab7daeb6460611

The final navigation took five mins to produce `page_navigation_settled`... I think `page_navigation_settled` should probably not depend on `network_idle`. In this case, one fetch to `https://ab.chatgpt.com/v1/initialize?...` never produced `loadingFinished`/`loadingFailed`, so `network_idle` did not fire until the 5-minute pending-request TTL swept it. That made `page_navigation_settled` arrive five minutes after the page was visibly usable.

Maybe we define `page_navigation_settled` around page lifecycle + visual stability, e.g. `page_load` plus `page_layout_settled`, and keep `network_idle` as a separate event. 


The other thing I noticed is that SPA/client-side route changes did not trigger events (on kernel.sh i clicked into the pricing page to simulate this). I think it's fine to treat inferring these as a follow-up because correctly distinguishing real route changes from prefetch/background fetches is harder.

archandatta:
Here is the latest run -> https://gist.github.com/archandatta/b7a841c0837afa27780177680228c878. Updated to not wait `network_idle`. Noted the SPA/client-side route changes issue will look into for the next pass!

rgarcia (approved):

PR #222, opened 2026-04-15
fix: add delay between clipboard sync and paste keystroke

hiroTamada:
## Summary

- Adds an 80ms delay between sending the clipboard update and the Ctrl+V keystroke in the live view paste handler
- Fixes the "one behind" clipboard issue where pasting consistently uses stale content

## Details

The clipboard update travels via WebSocket while the Ctrl+V keystroke travels via the WebRTC data channel. Even though the clipboard is sent first, the neko server processes key injection (X event queue) faster than the X11 clipboard selection write. Without the delay, the remote Chromium always pastes the *previous* clipboard content.

The 80ms delay gives the server time to complete the Xorg clipboard write before the paste keystroke arrives. This is a client-side workaround; the ideal fix would be server-side clipboard write acknowledgment.

Follow-up to #219.

## Test plan

- [ ] Open the live view inside a cross-origin iframe in Safari
- [ ] Copy text, Cmd+V into the live view — should paste correct text on the first attempt
- [ ] Copy different text, Cmd+V again — should paste the new text immediately (no "one behind")
- [ ] Verify right-click paste still works
- [ ] Test in Chrome to ensure no regressions

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Low Risk**
> Small, localized client-side timing change; main risk is introducing a slight paste delay or masking underlying server ordering issues.
> 
> **Overview**
> Adds an 80ms delay in `video.vue`’s `onPaste` handler between `syncClipboard()` and sending the remote Ctrl+V key sequence, to avoid pasting stale (“one behind”) clipboard content due to server-side X11 selection write latency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c1ce3a8b250992193262ba18856db53201fb629. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #223, opened 2026-04-15
fix: use paste event clipboardData for reliable first-paste in Safari iframes

hiroTamada:
## Summary

- Read clipboard text directly from `event.clipboardData.getData('text/plain')` in `onPaste`, which works synchronously in every browser — including Safari cross-origin iframes that block `navigator.clipboard.readText()` when the parent page lacks a `Permissions-Policy` header.
- Fall back to `navigator.clipboard.readText()` (via `syncClipboard()`) only when `clipboardData` is unavailable.
- Add an 80ms delay before sending the Ctrl+V keystroke so the neko server has time to write the clipboard to the Xorg selection.

Supersedes #222. Fixes the "first paste doesn't work in Safari through the dashboard" issue (CUS-163 / KERNEL-459).

## Root Cause

The dashboard embeds the live view in a cross-origin iframe with `allow="clipboard-read; clipboard-write"` but does **not** send the `Permissions-Policy: clipboard-read=*, clipboard-write=*` HTTP header. Safari blocks `navigator.clipboard.readText()` in this configuration, so `syncClipboard()` silently errors and no clipboard text reaches the remote on the first paste.

`event.clipboardData` is always populated by the browser during a native paste event regardless of iframe permissions, making it the reliable primary clipboard source.

## Test plan

- [x] Safari: Open dashboard → live view iframe → copy text → Cmd+V → text should paste on first attempt
- [x] Chrome: Same flow, verify no regression
- [ ] Firefox: Same flow, verify no regression
- [x] Verify subsequent pastes with new clipboard content also work without "one behind" lag

---

> [!NOTE]
> **Medium Risk**
> Touches clipboard and input-simulation timing; regressions could affect paste behavior across browsers or introduce subtle latency/race conditions in remote control.
> 
> **Overview**
> Improves remote paste reliability by changing `video.vue`’s `onPaste` handler to prefer synchronous clipboard reads from the native paste event (`event.clipboardData`) and only fall back to `syncClipboard()` when unavailable.
> 
> Adds a short (~80ms) delay before sending the Ctrl+V key sequence so the remote clipboard update can land before the paste keystroke, reducing “stale/one-behind” pastes (notably in Safari cross-origin iframes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff019f69be7f52919dda67ae429203bb7ec317cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/images/chromium-headful/client/src/components/video.vue
+++ b/images/chromium-headful/client/src/components/video.vue
@@ -845,17 +845,32 @@
       this.focused = false
     }
 
-    async onPaste() {
+    async onPaste(event: ClipboardEvent) {
       if (!this.hosting || this.locked) return
 
       try {
-        await this.syncClipboard()
+        // Read clipboard text directly from the paste event. This is
+        // synchronous and works in every browser — including Safari
+        // cross-origin iframes that block navigator.clipboard.readText()
+        // when the parent page lacks a Permissions-Policy header.
+        const text = event.clipboardData?.getData('text/plain')
+
+        if (text && text !== this.clipboard) {
+          this.$accessor.remote.setClipboard(text)
+          this.$accessor.remote.sendClipboard(text)
+        } else if (!text) {
+          await this.syncClipboard()
+        }
+
+        // Give the neko server time to write the clipboard text to the Xorg
+        // selection before we trigger the paste. The clipboard update arrives
+        // via WebSocket while the keystroke travels the WebRTC data channel;
+        // without this delay the remote pastes stale content.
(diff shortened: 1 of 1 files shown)


masnwilliams (approved):
lgtm — reading from event.clipboardData as the primary source is the right call for Safari iframe compat. fallback to syncClipboard and the 80ms delay are reasonable pragmatic choices. clean fix.

PR #217, opened 2026-04-14
Plan to add external events plan

archandatta:
> [!NOTE]
> **Low Risk**
> Low risk because this PR only adds a planning document and does not change runtime code or APIs.
> 
> **Overview**
> Adds `plans/external-events.md`, a detailed plan proposing new endpoints `POST /events/publish` (ingest external `events.Event` into the active capture session with enforced `source.kind`) and `GET /events/stream` (SSE streaming with `Last-Event-ID` resume, keepalives, and multi-client readers).
> 
> The plan also outlines follow-up behavioral changes like clearing capture-session state on stop via a `session_ended` marker, adding `CaptureSession.Active()`, and documenting both endpoints in `server/openapi.yaml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a06e95c09775d107b0d4554cdd8fbf49bba3ba7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #227, opened 2026-04-24
[kernel-1116] browser events: add external events

archandatta:
As per plan -> https://github.com/kernel/kernel-images/pull/217

---

> [!NOTE]
> **Medium Risk**
> Introduces new public event ingestion/streaming endpoints and refactors the core event/capture-session pipeline (seq semantics and envelope shape), which could impact existing consumers and event durability/ordering behavior.
> 
> **Overview**
> Adds **external event support** via new API endpoints: `POST /events/publish` to inject validated events into the server’s event bus and `GET /events/stream` to consume envelopes over SSE with `Last-Event-ID` resume and keepalives.
> 
> Refactors the event pipeline by splitting responsibilities into a process-wide `events.EventStream` (ring buffer + process-monotonic sequence) and a new `capturesession.CaptureSession` (session lifecycle + category filtering + stamping `capture_session_id` into `event.source.metadata`), removing the old `events.CaptureSession` and file-writing path. The OpenAPI-generated types/handlers are updated accordingly (new schemas, renamed capture category enums, SSE responses now set `Cache-Control: no-cache` and `X-Accel-Buffering: no`), along with build tooling tweaks to patch the additional SSE method.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb38112a0633af8f3e177f5b832e8a4b9972f7b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/lib/events/capturesession.go
```diff
@@ -172,12 +187,29 @@ func (s *CaptureSession) UpdateConfig(cfg CaptureConfig) {
...
-// is intentionally left intact so existing readers can finish draining.
-// A new session can be started by calling Start again.
+// Active reports whether a capture session is currently running.
+func (s *CaptureSession) Active() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.captureSessionID != ""
+}
+
+// Stop ends the current session. It publishes a synthetic session_ended
```
Sayan-: should `seq` be monotonic across start/stop cycles? If so, I think `Start` shouldn't reset it here. That would make `seq` represent the process-wide event stream position rather than the position within a single capture session.

server/cmd/api/api/events.go
```diff
@@ -0,0 +1,164 @@
...
+	"time"
+
+	"github.com/kernel/kernel-images/server/lib/events"
+	oapi "github.com/kernel/kernel-images/server/lib/oapi"
+)
+
+// PublishEvent handles POST /events/publish.
+// Injects a caller-supplied event into the active capture session. Returns 400
+// if no session is active or the event fails validation.
+func (s *ApiService) PublishEvent(_ context.Context, req oapi.PublishEventRequestObject) (oapi.PublishEventResponseObject, error) {
```
Sayan-, line 21: could we add one lifecycle test for the new event flow? Something like start capture session, publish an event through `PublishEvent`, read it from `StreamEvents`, then stop and verify the stream receives `session_ended`.

Sayan- (commented):
matches plan well and deviations from it are reasonable!

server/lib/events/capturesession.go
archandatta: I see your point, here are things to consider with how this works currently 
- every envelope already carries capture_session_id, so i think (session_id, seq) would be sufficient
- SSE clients don't cross session boundaries, session_ended terminates the stream, so not sure if monotonic seq applies fully

the fix would also require a non-trivial ring buffer refactor to avoid dropped records

server/cmd/api/api/events.go
archandatta: [4cb8f0e](https://github.com/kernel/kernel-images/pull/227/commits/4cb8f0e7118dfad8291705c03d8ab48b4c0415e6)

Sayan- (approved):
the implementation is internally leaning toward “seq is per capture session” but the docs don’t clearly say that. would recommend:

- make seq process-wide by not resetting it, or
- keep the current behavior and document that seq / Last-Event-ID are scoped to the active capture session

server/openapi.yaml
```diff
@@ -1274,8 +1274,119 @@ paths:
...
         "404":
           $ref: "#/components/responses/NotFoundError"
+  /events/publish:
+    post:
+      summary: Publish an event into the active capture session
+      description: >
+        Injects an event into the active capture session's event stream.
+        The event is assigned a monotonically increasing sequence number and
+        fanned out to all open stream connections. Returns 400 if no capture
+        session is active.
```
rgarcia: I would drop the "Returns 400 if no capture session is active" precondition and commit to a global event stream model: publish should always succeed and return the assigned envelope. The capture session is a native event publisher that overlays on top of the global event stream

server/openapi.yaml
```diff
@@ -1274,8 +1274,119 @@ paths:
...
+      operationId: publishEvent
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Event"
```
rgarcia: I wouldn't `$ref: Event` for the request body. The server forces `source.kind = kernel_api` and silently defaults to `system` category. I think this source.kind should be reserved for when we produce events on API requests to the image server and that we should let event publishers define whatever source.kind/category they want. I would define `PublishEventRequest` with only the fields the server actually honors so SDK consumers don't see fields that get overwritten.

server/openapi.yaml
```diff
@@ -1274,8 +1274,119 @@ paths:
...
+      responses:
+        "200":
+          description: Event accepted and published
```
rgarcia, line 1293: Empty 200 means callers can't correlate their POST to the stream. Return the assigned envelope

server/openapi.yaml
```diff
@@ -1274,8 +1274,119 @@ paths:
...
+          name: Last-Event-ID
+          schema:
+            type: string
+          required: false
+          description: >
+            Resume after this sequence number within the current capture
+            session. Omit or send 0 to start from the oldest buffered event.
+            Sequence numbers reset to 0 when a new capture session is started;
+            a value from a previous session is not meaningful here and may
+            cause the stream to miss events silently.
```
rgarcia: Under the global-stream model, seqs are process-monotonic — they don't reset on session start. This warning ("a value from a previous session is not meaningful") becomes obsolete and the silent-miss footgun goes away.

server/openapi.yaml
```diff
@@ -1274,8 +1274,119 @@ paths:
...
+          headers:
+            X-SSE-Content-Type:
+              description: Media type of SSE data events (application/json)
+              schema:
+                type: string
+                const: application/json
+          content:
+            text/event-stream:
+              schema:
+                type: string
```
rgarcia: is there no way to type this? what do we do in other sse streams here and in the API?

server/lib/events/capturesession.go
```diff
@@ -57,7 +57,9 @@ func NewCaptureSession(cfg CaptureSessionConfig) (*CaptureSession, error) {
 }
 
 // Start sets the capture session ID and applies the given config. It resets
-// the sequence counter so each session starts at seq 1.
+// the sequence counter so each session starts at seq 1. Sequence numbers are
+// scoped to the active session; Last-Event-ID values from a previous session
+// are not valid for reconnecting to a new one.
```
rgarcia: i would hoist ring, seq, and the system-event publish responsibility out of CaptureSession into a process-lifetime EventStream (or EventBus). CaptureSession then holds the stream as a dependency and is just (a) category filter for file writes + (b) capture-preferences metadata. This is the structural change that makes the global-stream model real.

we may also want to just rm the file writing since in a global event stream model it is less clear what events go to these files

rgarcia (changes requested):

server/cmd/api/api/events.go
```diff
@@ -32,9 +32,7 @@ func (s *ApiService) PublishEvent(_ context.Context, req oapi.PublishEventReques
 
 	ev := events.Event{Type: body.Type}
 
-	if body.Ts != nil {
-		ev.Ts = *body.Ts
-	}
+	ev.Ts = time.Now().UnixMicro()
```
archandatta, line 28: @raf any thoughts for this? I'm going back and forth if we should allow for the timestamp to be sent or not because then the request could completely miss align with the event

server/lib/events/capturesession.go
archandatta: Adressed -> [5362a27](https://github.com/kernel/kernel-images/pull/227/commits/5362a272406d01565f59fed7a11eb2e1b87a8c49)

server/openapi.yaml
archandatta: good catch i missed updating this after creating the event structs [56782c0](https://github.com/kernel/kernel-images/pull/227/commits/56782c01759bdda3c286985656a9e56520fe631d)

server/openapi.yaml
archandatta: addressed -> [7bc3050](https://github.com/kernel/kernel-images/pull/227/commits/7bc3050054b547172db2be58285a083a75f2229b)

server/openapi.yaml
archandatta: addressed -> [059f929](https://github.com/kernel/kernel-images/pull/227/commits/059f929836f28259962f472fb2438117a6dcfb67)

server/openapi.yaml
archandatta: addressed -> [093d8cc](https://github.com/kernel/kernel-images/pull/227/commits/093d8cc336d9ad902a08a74fdac2041f305990b7)

rgarcia (commented):
some things still outstanding in terms of decoupling the publish/subscribe API from the capture session:
- PublishEvent 400s when no capture session active (events.go:21-23)
- StreamEvents 400s when no capture session active (events.go:82-84)
- SSE stream terminates on `session_ended` (events.go:145-147) 

Also nit but we should rename the capture session events to capture_session otherwise we risk confusion with the browser session itself beginning/ending 

External publish should be treated like a peer producer to the cdp producer, so I think instead of `Publish` and `PublishUnfiltered` living on CaptureSession it should be something like

```
// EventStream gets a public Publish (today's es.publish is private)
func (es *EventStream) Publish(env Envelope) Envelope

// CaptureSession.Publish stays, applies its own config-driven filter,
// internally calls es.Publish
func (s *CaptureSession) Publish(ev Event)

// PublishUnfiltered goes away
```

Also it looks like capture session id is still a top level field on the event envelope. 

```
// today
type Envelope struct {
    CaptureSessionID string  // producer detail leaking up
    Seq              uint64  // actual pipeline metadata
    Event            Event
}
```

In a global-event stream model I think it should be

```
type Envelope struct {
    Seq   uint64
    Event Event
}

type Source struct {
    Kind             SourceKind
    Event            string
    Metadata         map[string]string // capture session id can go here as a producer detail
}
```

also the SSE stream data and response to POST /events should be identical to the envelope i.e. `{seq: ..., event: ...}`

rgarcia (commented):
- let's drop events_dropped and capture_session_ended as synthetic frames on the stream. Both can be inferred by clients without server-side synthesis. Removing them simplifies the wire format (one frame shape, no seq=0 sentinel, no omitted-id: carve-out) and stops mixing system control signals with user payload events.

- `PublishedEnvelope.seq` openapi.yaml description still says "assigned to this event within the current capture session" — stale now that seqs are process-monotonic and the stream isn't session-scoped. Reword to "process-monotonic across the lifetime of the server."

rgarcia (approved):
LGTM — global event stream model is fully realized. Bus is decoupled from session lifecycle, capture session is one producer among many, schemas are clean across POST and SSE, and the wire format is uniform now that the synthetic control frames are gone. Drop-detection contract is documented for SDK consumers. Ship it.

PR #230, opened 2026-04-27
chromium-launcher: instrument restart phases to localise slow boots

hiroTamada:
## Why

When chromium is restarted (e.g. after a flag/policy change or a profile load), callers need DevTools to come back up within a tight budget. In our downstream production deployment we see that some restarts take ~6 s and exceed that budget — causing user-visible failures — while others complete in ~1 s on the very same fleet. Restarts are clearly **bimodal**, but right now we cannot tell which phase eats the time on the slow ones.

The plausible suspects are:

- `supervisorctl restart chromium` itself (SIGKILL + spawn) being slow on a degraded host.
- `chromium-launcher`'s `killExistingChromium` (`pkill -9` + `pgrep` wait) or `waitForPort`. We disable `SO_REUSEADDR`, so any lingering FDs / TIME_WAIT entries on the devtools port force us to wait for kernel cleanup — up to the 5 s `waitForPort` deadline.
- Chromium's internal init after `syscall.Exec` (cache warmup, profile work) running slowly post-suspend / post-standby-restore.

This PR adds the instrumentation we need to tell those apart.

## What

Two structured events. Both are correlatable downstream by host / VM identity, so they can be joined with each other and with caller-side CDP-readiness timings to get an end-to-end phase breakdown.

### 1. `chromium_launcher_phase_complete` (in `chromium-launcher`)

Emitted as slog JSON to stderr right before `syscall.Exec`. Captures:

- **Phase timings**: `pkill_elapsed_ms`, `pgrep_wait_elapsed_ms`, `pgrep_attempts`, `pgrep_timed_out`, `waitport_elapsed_ms`, `waitport_attempts`, `waitport_timed_out`, `waitport_last_err`, `launcher_total_ms`.
- **Devtools-port socket state** parsed from `/proc/net/tcp`: `ss_estab_devtools`, `ss_timewait_devtools`, `ss_listen_devtools` — directly tests the lingering-FD / TIME_WAIT hypothesis.
- **VM state**: `vm_uptime_ms`, `load_avg_1m`, `mem_available_kb`, `mem_cached_kb`, `mem_dirty_kb` — directly tests the post-standby-restore / host-pressure hypothesis.
- `instance_name` (read from `INSTANCE_NAME` env var if set) for downstream attribution.

`killExistingChromium` and `waitForPort` are refactored to return small phase-timing structs (\`killStats\`, \`waitportStats\`); behavior is unchanged.

### 2. `chromium_restart_complete` (in \`server/cmd/api/api/chromium.go\`)

Emitted from \`restartChromiumAndWait\` after a successful restart. Decomposes the wall clock into:

- \`supervisorctl_ms\` — time for \`supervisorctl restart chromium\` to return.
- \`post_supervisorctl_ms\` — gap between supervisorctl returning and a new DevTools URL appearing in the supervisord log tail.
- \`total_ms\`, \`operation\`, \`first_devtools_url_len\`.

This separates "supervisord was slow to SIGKILL+respawn" from "chromium itself was slow to initialise."

Combined with the launcher event we get: \`supervisorctl_ms\` → \`launcher_total_ms\` (with sub-phases) → chromium internal init → DevTools URL written.

### \`/proc\` parsers

New helpers \`readVMUptimeMs\`, \`readLoadAvg1m\`, \`parseMemInfo\`, \`countSocketStates\` are all fail-soft (return zero on missing files / parse errors) and have unit tests in \`main_test.go\` with realistic \`/proc/net/tcp\` and \`/proc/meminfo\` fixtures (using testify \`assert\` / \`require\`).

## Test plan

- [x] \`go build ./...\` (server)
- [x] \`go vet ./...\`
- [x] \`go test ./cmd/chromium-launcher/...\` — passes (includes new fixture-based tests)
- [x] \`go test ./cmd/api/...\` — passes
- [x] \`go test ./lib/devtoolsproxy/...\` \`./lib/chromiumflags/...\` — passes
- [ ] After deploy, verify both events show up downstream with expected fields.
- [ ] Pivot \`chromium_restart_complete.total_ms\` by host to confirm the bimodal distribution and check correlation with \`vm_uptime_ms\` / \`ss_timewait_devtools\` from the launcher event.

## Notes

- Pure observability — no behavior changes to the restart path. Existing call sites of \`killExistingChromium()\` and \`waitForPort(...)\` were updated to discard the new return values where appropriate; nothing in main \`exec\`'s shape changed.
- The new event is written to stderr where the existing diagnostic prints already go, so it'll be picked up by the same log pipeline as everything else \`chromium-launcher\` emits.

cc @<reviewer>

---

> [!NOTE]
> **Medium Risk**
> Touches the Chromium restart/wait control flow and launcher startup path; while intended to be observability-only, mistakes could change restart timing or logging behavior and affect readiness detection.
> 
> **Overview**
> Adds **structured phase instrumentation** for Chromium restarts so slow boots can be attributed to specific steps.
> 
> `restartChromiumAndWait` now measures `supervisorctl restart` duration and the delay until a new DevTools URL appears, emitting `chromium_restart_complete` (and `chromium_restart_timeout` with partial timing on timeout).
> 
> `chromium-launcher` now records timing for `pkill`/`pgrep` and `waitForPort`, plus fail-soft VM/socket telemetry from `/proc` (uptime, load, meminfo, `/proc/net/tcp` state counts), emitting a single JSON `chromium_launcher_phase_complete` event immediately before `exec`ing Chromium; new unit tests cover the `/proc` parsers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67244b58b921ab7f45dceecc0aa4801eac84e448. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/cmd/api/api/chromium.go
+++ b/server/cmd/api/api/chromium.go
@@ -333,6 +333,11 @@ func writeChromiumFlags(tokens []string) error {
 
 // restartChromiumAndWait restarts Chromium via supervisorctl and waits for DevTools to be ready.
 // Returns an error if the restart fails or times out.
+//
+// On success, emits a chromium_restart_complete log line decomposing the wall clock into
+// supervisorctl_ms (time for supervisorctl restart to return) and post_supervisorctl_ms
+// (gap between supervisorctl returning and a new DevTools URL appearing). Combined with
+// launcher_total_ms from the chromium-launcher event, this localises slow restarts.
 func (s *ApiService) restartChromiumAndWait(ctx context.Context, operation string) error {
 	log := logger.FromContext(ctx)
 	start := time.Now()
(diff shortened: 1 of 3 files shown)


rgarcia (approved):
lgtm as long as you've given it some e2d test reps

server/cmd/api/api/chromium.go
```diff
@@ -341,32 +346,61 @@ func (s *ApiService) restartChromiumAndWait(ctx context.Context, operation strin
...
-		log.Info("devtools not ready in time", "operation", operation, "elapsed", time.Since(start).String())
-		return fmt.Errorf("devtools not ready in time")
+	for {
+		select {
+		case r := <-resultCh:
+			if r.err != nil {
+				return fmt.Errorf("supervisorctl restart failed: %w", r.err)
+			}
+			supervisorctlMs = r.elapsedMs
+			supervisorctlDoneAt = time.Now()
```
Sayan-, line 380: `supervisorctlDoneAt` is set when this goroutine *receives* from resultCh, not when the launcher goroutine finished supervisorctl. Under scheduling delay — i.e. the slow-restart case this PR exists to debug — `post_supervisorctl_ms` will be biased low and won't sum to `total_ms`. Record `time.Now()` inside the goroutine and pass it through `restartResult`.

server/cmd/api/api/chromium.go
```diff
@@ -341,32 +346,61 @@ func (s *ApiService) restartChromiumAndWait(ctx context.Context, operation strin
...
+			// Drop resultCh from select so we keep waiting for updates without re-firing.
+			resultCh = nil
+		case url := <-updates:
+			postSupervisorMs := int64(-1)
```
Sayan-, line 384: When `updates` wins the race vs. `resultCh`, both `supervisorctl_ms` and `post_supervisorctl_ms` log as -1. Downstream aggregates (medians, p99s) need to filter `>= 0` or they'll be poisoned by sentinels. Worth a comment here and a note wherever the query lives.

server/cmd/api/api/chromium.go
```diff
@@ -341,32 +346,61 @@ func (s *ApiService) restartChromiumAndWait(ctx context.Context, operation strin
...
-	select {
-	case <-updates:
-		log.Info("devtools ready", "operation", operation, "elapsed", time.Since(start).String())
-		return nil
-	case err := <-errCh:
-		return err
-	case <-timeout.C:
-		log.Info("devtools not ready in time", "operation", operation, "elapsed", time.Since(start).String())
-		return fmt.Errorf("devtools not ready in time")
+	for {
```
Sayan-, line 373: The new for-select + channel-nilling pattern is the highest-behavior-risk diff in the PR and has no test. The `/proc` parsers got fixtures; the actual control flow did not. Could you add a test covering: (a) supervisorctl succeeds then updates arrives, (b) updates arrives before supervisorctl returns, (c) supervisorctl errors? The PR description says "no behavior changes" but the loop shape did change — would be good to lock in.

Sayan- (commented):
few notes about timing precision. overall 🚀

PR #234, opened 2026-05-02
Rewrite chromium wrapper.sh as a Go binary

sjmiller609:
## Summary

Replaces the bash `wrapper.sh` shipped in `chromium-headful` and `chromium-headless` with a single Go binary (`server/cmd/wrapper`) that detects the profile at boot from supervisor's `conf.d` (`xorg.conf` → headful, `xvfb.conf` → headless).

The wrapper is split into two phases so identity-bound work doesn't block boot:

- **Phase A** (identity-free) starts `xorg`/`xvfb`, `dbus`, `chromedriver`, and `chromium` in one `supervisorctl` call, then probes readiness concurrently. Per-service files (`chromium.go`, `display.go`, `envoy.go`, `probes.go`, `supervisord.go`, `system.go`) keep each concern isolated.
- **Phase C** (identity-bound) renders the envoy bootstrap from `INST_NAME` / `JWT` / `XDS_SERVER` and brings up `kernel-images-api` + envoy once those are present.

X11 socket readiness is a small shared package (`server/lib/x11`) used by both the wrapper and the chromium launcher so they agree on what "display ready" means.

Per-service tweaks:

- supervisor confs: `startsecs=2` → `0` so `supervisorctl start` returns immediately; the Go wrapper probes readiness directly.
- `init-envoy.sh`: drop the trailing port poll / curl-through-proxy check — `waitAllReady` covers both, and the script now fails fast on any error.
- `Kraftfile` `cmd` updated from `/wrapper.sh` to `/wrapper`.
- Envoy proxy CA cert is baked at image build time.

Other behaviour:

- SIGTERM/SIGINT handler is registered before `supervisord` starts so signals during the seconds-long startup window are caught instead of killing the wrapper outright.
- Critical boot scripts (`init-envoy.sh`) are streamed via `runStreamFatal` so failures abort boot instead of being silently swallowed, matching the old `set -o errexit`.
- The ready log line reports per-probe durations, e.g. `ready in 2.6s (phaseA=251ms phaseC=289ms; cdp=2.646s chromedriver=1.468s forward-proxy=2.306s)`.

Cosmetic + non-critical work (`pulseaudio`, `--no-sandbox` infobar dismissal) runs off the hot path.

## Test plan

- [ ] `go build ./cmd/wrapper` and `go vet ./cmd/wrapper` pass
- [ ] Build chromium-headless image and confirm container reaches CDP-ready
- [ ] Build chromium-headful image and confirm container reaches CDP-ready
- [ ] Confirm envoy comes up and the forward-proxy listener is reachable when `INST_NAME` / `XDS_SERVER` / `KERNEL_INSTANCE_JWT` are set
- [ ] Confirm `RUN_AS_ROOT=true` headful flow still dismisses the `--no-sandbox` infobar
- [ ] Confirm SIGTERM/SIGINT propagates to supervisord and child services exit cleanly

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it replaces the container entrypoint/startup orchestration for both headful and headless Chromium images and changes Envoy certificate/bootstrap behavior, which could affect boot readiness and proxy trust in production.
> 
> **Overview**
> **Replaces `wrapper.sh` with a single Go entrypoint (`/wrapper`)** for both `chromium-headful` and `chromium-headless`, updating Dockerfiles, Kraft `cmd`, and README examples to use the new binary.
> 
> The new wrapper starts `supervisord` in foreground, brings up core services in parallel (X server, `dbus`, `chromedriver`, `chromium`), and performs explicit readiness probes (CDP, chromedriver, optional `neko`, optional Envoy) instead of relying on `supervisor` delays; corresponding supervisor `startsecs` values are set to `0`.
> 
> **Envoy setup is split**: a new build-time `shared/envoy/bake-certs.sh` bakes the forward-proxy CA into system trust + NSS DBs, while `shared/envoy/init-envoy.sh` now only renders bootstrap from identity env and starts/restarts Envoy (dropping runtime cert generation and inline readiness/proxy tests). `chromium-launcher` now waits for X display (and `mutter` when headful) using a new `server/lib/x11` helper to avoid restart loops/WM decoration issues.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae4b8d3b015261e40cc59a6cb938ef046dbe2137. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/cmd/wrapper/main.go
```diff
@@ -0,0 +1,330 @@
...
+		phaseCDone.Sub(phaseCStart).Truncate(time.Millisecond),
+		formatProbeDurations(probeDurations))
+
+	// Cosmetic + non-critical services come up off the hot path. Headless has
+	// no audio stack and no UI to dismiss.
+	if prof == profileHeadful {
+		go func() {
+			startAll("pulseaudio")
+			if os.Getenv("RUN_AS_ROOT") == "true" {
+				dismissNoSandboxWarning()
```
sjmiller609: is this acceptable async from hot path?

images/chromium-headful/supervisor/services/chromedriver.conf
```diff
@@ -2,6 +2,6 @@
 command=/usr/local/bin/chromedriver --port=9225 --allowed-ips=127.0.0.1 --log-level=INFO
 autostart=false
 autorestart=true
-startsecs=2
```
sjmiller609, line 5: a lot of the time was just from waiting on sysctl start with 2 second hard code wait every time we run the cli, but we don't need that given our script also does probes to wait on readiness

server/cmd/wrapper/main.go
```diff
@@ -0,0 +1,331 @@
...
+	//        expansion in init-envoy.sh and the supervisorctl-spawned
+	//        services inherit them.
+	//     3. The identity phase uses `supervisorctl restart envoy`
+	//        (idempotent — start on first boot, stop+start on a re-render
+	//        after fork) so a restored snapshot drops its stale identity
+	//        cleanly.
+	//   Boot path keeps running through unchanged: the wait simply no-ops
+	//   when there's no fork bundle to receive.
+
+	// Identity phase: identity-bound services. Render envoy bootstrap with
```
sjmiller609, line 205: here and down will be hot path

server/cmd/wrapper/main.go
sjmiller609: we are deleting this, n/a any more

rgarcia (approved):

PR #237, opened 2026-05-11
API for out-of-band scale-to-zero pin

sjmiller609:
## Summary

Adds two new endpoints to the kernel-images server:

- `POST /system/standby/disable` — pins scale-to-zero off until released
- `POST /system/standby/enable` — releases the pin

The pin lives alongside the existing request-driven middleware refcount inside `DebouncedController`:

- scale-to-zero stays disabled while **either** request holders remain inflight **or** the pin is held
- request-driven `Enable` (from middleware) does **not** release the pin
- releasing the pin while requests are inflight defers the underlying enable until the last request completes
- the configured re-enable cooldown applies on pin release exactly as it does on request release

`DebouncedController` and `NoopController` now also implement a new `PinnedController` sub-interface (`Controller` + `DisablePin` / `EnablePin`). The pin is a boolean — `DisablePin` / `EnablePin` are idempotent.

## Why

This is the in-VM surface needed for a future control-plane integration: an external system (e.g. a hot-pool controller) needs to hold a VM out of standby while it sits idle in a pool, then release the hold when the VM is claimed.

The existing middleware refcount only works for inflight HTTP requests, so it can't hold a VM disabled across an idle period.

## Notes for reviewers

- All existing middleware-driven flows are byte-identical (the pin defaults to false; all current call sites take the same code paths).
- The constructor return type for `NewDebouncedController*` widened from `Controller` to `*DebouncedController` so callers can access the pin methods. Only existing caller is `cmd/api/main.go`, which is unaffected since `*DebouncedController` still satisfies `Controller` for `recorder.NewFFmpegRecorderFactory` and `scaletozero.Middleware`.
- Control-plane wiring (metro-api proxy + API server) is intentionally **not** in this PR.

## Test plan

- [x] `go test -race ./lib/scaletozero/...` passes (6 new tests covering pin semantics)
- [x] `go test -race ./cmd/api/...` passes
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] Manual review of openapi.yaml + regenerated `lib/oapi/oapi.go`

---

> [!NOTE]
> **Medium Risk**
> Introduces new API endpoints and modifies `DebouncedController` state machine to include a pinned hold, which could affect VM standby/idle behavior and timing if the pin logic misbehaves.
> 
> **Overview**
> Adds control-plane style endpoints `POST /scaletozero/disable` and `POST /scaletozero/enable` that *pin* scale-to-zero off/on independently of the existing per-request middleware refcount.
> 
> Extends `scaletozero` with a new `PinnedController` interface and updates `DebouncedController` to track a boolean `pinned` hold that blocks request-driven re-enables until explicitly unpinned (still honoring the existing cooldown behavior).
> 
> Regenerates `openapi.yaml`/`lib/oapi` for the new routes and adds unit + e2e coverage to validate idempotency and that the pin does not interfere with normal request handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e15d112b021e19062741fe679ad938bcae74b0a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/openapi.yaml
```diff
@@ -1334,6 +1334,33 @@ paths:
...
+          description: Standby pinned disabled
+        "500":
+          $ref: "#/components/responses/InternalError"
+  /system/standby/enable:
+    post:
+      summary: Release the standby pin set by /system/standby/disable
+      description: >
+        Releases the out-of-band scale-to-zero pin. If no request-driven
+        holders remain, scale-to-zero re-enables (honoring any configured
+        cooldown). Idempotent — calling without a held pin has no effect.
```
Sayan-: pls deslop descriptions. these read like internal implementation notes rather than information a caller needs: what does calling this endpoint do for me, when do I call it, and what pairs with it

server/openapi.yaml
```diff
@@ -1334,6 +1334,33 @@ paths:
             text/event-stream:
               schema:
                 $ref: "#/components/schemas/PublishedEnvelope"
+  /system/standby/disable:
```
Sayan-: would recommend making the path/operation more specific to `scaletozero` since the current naming could imply it's actually mutating the state of the vm, which is _not_ what we're doin

server/lib/scaletozero/scaletozero.go
```diff
@@ -20,6 +20,24 @@ type Controller interface {
...
+// reserving a VM in a hot pool) where the holder is not tied to an inflight
+// HTTP request.
+type PinnedController interface {
+	Controller
+	// DisablePin pins scale-to-zero disabled until EnablePin is called. The
+	// pin is a boolean, not a counter: repeated calls are idempotent.
+	DisablePin(ctx context.Context) error
+	// EnablePin releases the pin. If no request-driven holders remain,
+	// scale-to-zero is re-enabled (honoring any configured cooldown).
+	EnablePin(ctx context.Context) error
```
Sayan-: these operations seem inverted. I had to re-read this a few times to wrap my head around it. would recommend revisiting semantics here to simplify. even `Pin / Unpin` would be clearer

Sayan- (changes requested):

server/openapi.yaml
sjmiller609: Rewrote both summary/description to be caller-focused — what the endpoint does, when to call it, and what pairs with it. See a41fed3.

server/openapi.yaml
sjmiller609: Renamed paths to `/scaletozero/pin` and `/scaletozero/unpin`, operationIds to `pinScaleToZero`/`unpinScaleToZero`. a41fed3.

server/lib/scaletozero/scaletozero.go
sjmiller609: Renamed to `Pin`/`Unpin` on the `PinnedController` interface, `DebouncedController`, and `NoopController`. a41fed3.

server/openapi.yaml
sjmiller609: hold on, I'm manually reviewing these now

server/openapi.yaml
```diff
@@ -1334,6 +1334,33 @@ paths:
             text/event-stream:
               schema:
                 $ref: "#/components/schemas/PublishedEnvelope"
+  /scaletozero/pin:
```
sjmiller609: I think the path like enable / disable is more clear

server/cmd/api/api/api.go
```diff
@@ -96,7 +96,7 @@ func New(
 	recordManager recorder.RecordManager,
 	factory recorder.FFmpegRecorderFactory,
 	upstreamMgr *devtoolsproxy.UpstreamManager,
-	stz scaletozero.Controller,
+	stz scaletozero.PinnedController,
```
sjmiller609, line 99: getting rid of "pin" terminology

sjmiller609: actually maybe Pin does make the most sense

Sayan- (approved):
yeet - thanks for iterating!

PR #252, opened 2026-05-26
Record replay audio

rgarcia:
## Summary

Adds **audio to replay recordings**, opt-in per recording via a new `recordAudio` flag on the recording API. Default is `false`, so existing callers are completely unchanged (video-only, identical encoding to `main`).

To make browser audio actually exist to capture, the browser images now run a shared PulseAudio stack:

- `shared/start-pulseaudio.sh` creates a null **sink** `KernelOutput` (Chromium plays into it; the recorder captures its `.monitor`) and a standalone null **source** `KernelInput` (so the browser advertises a real, non-monitor microphone — Chromium hides monitor sources from `enumerateDevices()`, which antibot scripts flag).
- Chromium routes playback into `KernelOutput`, ffmpeg captures `KernelOutput.monitor` when `recordAudio=true`, and Neko live view captures the same monitor so the stream has audio.
- PulseAudio is now a boot dependency for both headful and headless (previously headful-only, off the hot path).

## Design decisions & rationale

**Audio is opt-in per recording, not a VM-level toggle.** "Should this recording have audio" is a property of the recording request, so it lives on `POST /recording/start` (`recordAudio`, default `false`) rather than an env var. This keeps every existing caller on the original behavior and makes the new A/V-oriented encoding strictly opt-in. The control plane opts in where it wants audio (see Deploy ordering).

**Config location — one authority, per-consumer defaults, no ambient env.** The audio topology (socket path, sink/source names) is owned by `shared/start-pulseaudio.sh` — it's the producer that actually creates them. Every consumer keeps its own matching default next to where it's used, with a comment pointing back to that authority:
- recorder (`kernel-images-api`): `AUDIO_SOURCE` / `PULSE_SERVER` are the "where to capture" address (the image's supervisor conf sets both; `recordAudio` is the separate "whether").
- `chromium-launcher`: hardcodes the `PULSE_SERVER` / `PULSE_SINK` contract values (it's the sole consumer of `PULSE_SINK`, and Chromium's `AudioManagerPulse` honors it to route playback into `KernelOutput`).
- Neko: `PULSE_SERVER` in its supervisor conf.

We deliberately avoided centralizing these in the container entrypoint (the wrapper) as ambient env. The wrapper isn't the producer or a consumer, and ambient seeding created a fragile ordering dependency (consumers silently getting empty values). Duplicating a handful of extremely stable string constants across a bash producer + Go binaries + supervisor INIs is the lesser evil, made safe with cross-reference comments. The wrapper's old `PULSE_*` seeding block was removed for this reason.

**`recordAudio` gates the encoder, fixing a video-only regression.** `-use_wallclock_as_timestamps` (added originally for stable video-only playback) is kept for video-only capture but dropped when recording audio — with two inputs (x11grab + pulse) it stamps each independently and desyncs them. Gating it on `!recordAudio` makes the default path byte-identical to `main` while keeping A/V aligned when audio is on. `-preset veryfast` / `-tune zerolatency` / AAC likewise apply only when recording audio.

**Validation moved to where it can fail fast.** `recordAudio=true` with no source/socket configured is rejected at recorder creation (the factory validates merged params), instead of producing a broken ffmpeg invocation.

**Startup hardening (`start-pulseaudio.sh`):** the unix socket module loads *last*, so the socket the wrapper waits on only appears once `KernelOutput` is registered (closes a socket-ready-before-sink race). The silent keepalive stream is self-healing and the script blocks only on the daemon, so a `pacat` hiccup no longer tears down the whole audio stack.

**Headless `--mute-audio` removed** so headless sessions can produce capturable audio.

## Deploy ordering / follow-up

This is the kernel-images side and is backwards-compatible (default `false`). Because of that, **replays stay video-only until the consumer opts in**: a separate `packages/api` PR must pass `recordAudio=true` from the `record_session` replay-capture flow. Ship kernel-images first, then the consumer.

## Tests

- `go test ./lib/recorder ./cmd/config ./cmd/api/... -count=1` (unit: ffmpeg audio args, the `recordAudio` validation matrix, and wall-clock gating)
- `DOCKER_BUILDKIT=1 docker build -f images/chromium-headful/Dockerfile -t onkernel/chromium-headful-test:latest .`
- `E2E_CHROMIUM_HEADFUL_IMAGE= go test ./e2e -run TestReplayRecordingIncludesAudioTrack -count=1 -v` — records browser audio with `recordAudio=true`, downloads the MP4, asserts a non-silent audio track whose duration reaches the end of the recording, and verifies the browser sees a non-monitor mic over CDP. (`TestReplayRecordingZombocomArchiveAudio` is opt-in via `RECORDING_ZOMBO_OUTPUT_PATH`.)

---

> [!NOTE]
> **Medium Risk**
> Touches container boot order, PulseAudio topology, and ffmpeg encoding paths; default-off API limits blast radius, but mis-synced constants or startup races could break audio or live view.
> 
> **Overview**
> Adds **opt-in replay audio** via `recordAudio` on `POST /recording/start` (default **false**, so existing callers stay video-only). When enabled, ffmpeg muxes Pulse capture from `KernelOutput.monitor` alongside X11 video, with validation that returns **400** if audio is requested without `AUDIO_SOURCE` / `PULSE_SERVER`.
> 
> Headful and headless images now share **`shared/start-pulseaudio.sh`**: a `KernelOutput` null sink for browser playback and recording, a `KernelInput` null source so `enumerateDevices()` shows a real mic, and a unix socket at `/tmp/pulse/native`. Chromium is wired with `PULSE_SERVER` / `PULSE_SINK`; the wrapper starts **pulseaudio before chromium** and waits on the socket. Supervisor passes audio env into **kernel-images-api** and **neko** (`KernelOutput.monitor` in `neko.yaml`). Headless drops **`--mute-audio`**; the Neko client unmutes on keyboard/mouse interaction.
> 
> E2E tests assert non-silent MP4 audio tracks and CDP-visible input/output devices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9538d6681b0407f51830fe06a96ba0ff44e13ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/lib/recorder/ffmpeg.go
```diff
@@ -472,44 +504,83 @@ func (fr *FFmpegRecorder) Delete(ctx context.Context) error {
...
+			"-ar", "48000",
+			"-ac", "2",
+		}...)
+	}
+
+	args = append(args, []string{
 		// Timestamp handling for reliable playback
-		"-use_wallclock_as_timestamps", "1", // Use system time instead of input stream time
-		"-reset_timestamps", "1", // Reset timestamps to start from zero
+		"-reset_timestamps", "1",
```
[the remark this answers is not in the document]
rgarcia: going to change the recording api to have a boolean `recordAudio` that is false by default to reduce risk here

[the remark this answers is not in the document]
rgarcia: b8f1f8e

server/cmd/wrapper/main.go
```diff
@@ -172,11 +175,13 @@ func main() {
 	_ = os.WriteFile(filepath.Join(supervisordLogD, "chromium"), nil, 0o644)
 
 	browserStart := time.Now()
-	startAll(xServer, "dbus", "chromedriver", "chromium")
+	startAll(xServer, "dbus", "chromedriver", "pulseaudio")
 	waitForX(defaultDisplay, 20*time.Second)
 	if prof == profileHeadful {
 		startAll("mutter")
 	}
+	waitForSocket(pulseSocket, 10*time.Second)
```
hiroTamada, line 183: Just curious how long this adds in terms of the startup time - not that it matters with fork

shared/start-pulseaudio.sh
```diff
@@ -0,0 +1,92 @@
...
+    sleep 0.1
+  done
+
+  # Keep a silent stream open on the sink so its monitor always produces data for
+  # the recorder. Self-heal if pacat dies: a keepalive hiccup should not tear down
+  # the whole audio stack, so restart it as long as the daemon is alive and only
+  # let the script exit when pulseaudio itself does.
+  (
+    while kill -0 "$pulse_pid" 2>/dev/null; do
+      pacat --raw --rate=48000 --channels=2 --format=s16le --device="$KERNEL_SINK" /dev/zero || true
```
hiroTamada, line 85: Just thinking ahead. Is `--rate=48000` something we want to dynamically change in the future?

hiroTamada (approved):
LGTM. Nothing to flag from my side

server/cmd/wrapper/main.go
rgarcia: Measured on a cold headful boot (v6 image): the pulse socket (`/tmp/pulse/native`) becomes ready ~ 60ms after the X display socket (`/tmp/.X11-unix/X1`) — 854.579 vs 854.519. Since the wrapper waits for X (and headful waits for mutter) *before* the `waitForSocket(pulseSocket)` gate, the marginal latency added before chromium starts is ≤~60ms, and it overlaps mutter startup — effectively negligible against the ~0.8–1s browser bring-up (chromedriver alone was ready at 811ms in the same boot). And as you said, snapshot/fork restore skips cold boot entirely.

shared/start-pulseaudio.sh
rgarcia: 48kHz isn't really a choice here — it's the standard rate the whole pipeline already assumes end-to-end: WebRTC/Opus for the neko live view and the AAC encode in the recorder (`-ar 48000`). The null sink/source and the silent keepalive all use it so every stage stays sample-rate-aligned with no resampling. No current driver to vary it; if one ever came up we'd thread a single rate constant through this script + the recorder's `-ar`, but I'd hold off until there's a real need.

PR #264, opened 2026-05-30
Surface display client errors

sjmiller609:
## Summary
- preserve Neko screen configuration status codes so display 4xx responses stay client-visible
- return 400s from batched display configure when Neko rejects the requested screen mode
- log requested display dimensions, refresh rate, display mode, and available runtime identity metadata on display configure failures

## Tests
- go test ./cmd/... ./lib/...


sjmiller609:
Superseded by an API-only fix in kernel/kernel.

PR #265, opened 2026-06-02
Decode CHROMIUM_FLAGS_B64 in chromium-launcher

yummybomb:
## Summary
- Some unikernel platform versions backslash-escape spaces when serializing env vars into the guest kernel command line, corrupting the space-separated `CHROMIUM_FLAGS`. The launcher splits on whitespace, so `--no-sandbox` becomes `--no-sandbox\`, Chromium ignores it, aborts with `No usable sandbox`, and supervisord crash-loops it.
- `chromium-launcher` now prefers a base64url-encoded `CHROMIUM_FLAGS_B64`, decoding it into the base flag string. base64url has no characters a cmdline encoder escapes, so it survives transport intact.
- Falls back to `CHROMIUM_FLAGS` when the encoded var is unset or malformed, so behavior is unchanged with an older API.

## Backward compatibility / idempotency
- Decode is a pure read with no write-back; every launcher start (including supervisord restarts and scale-to-zero restores) resolves the same flags.
- `CHROMIUM_FLAGS_B64` unset → identical to previous behavior. Malformed value → logs and falls back to the raw value.

## Deploy ordering — this PR ships FIRST
The companion API PR (`kernel/kernel`) switches to sending **only** `CHROMIUM_FLAGS_B64` (sending both overflows Firecracker's kernel cmdline limit on the longest variant). So this launcher change must land and every browser image variant must be rebuilt and rolled out **before** the API PR deploys — otherwise headful/stealth browsers would get no base flags. This PR is backward-compatible with the current API (raw-only) via the fallback, so it is safe to ship and roll out ahead of time.

## Test plan
- [x] `go test ./cmd/chromium-launcher/ ./lib/chromiumflags/` — `TestResolveBaseFlags` covers precedence, blank/malformed fallback, and round-trip of a value containing spaces, a quoted token, and comma lists.
- [ ] Build a headless image from this branch, provision on an affected host with `CHROMIUM_FLAGS_B64`, and confirm `FINAL_FLAGS` shows a clean `--no-sandbox` and DevTools comes up.

### DIFF
--- a/server/cmd/chromium-launcher/main.go
+++ b/server/cmd/chromium-launcher/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/base64"
 	"flag"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


yummybomb:
Closing — exploring a wrapper-level fix in cmd/wrapper instead of the launcher b64 decode.

PR #272, opened 2026-06-03
Telemetry categories: opt-in selection + lightweight default set

Sayan-:
### What changes

- Category selection is now opt-in. A `browser` config captures exactly the categories set to `enabled: true`; anything omitted is off (previously omitted = on). A request that enables nothing clears telemetry.
- The default set (no `browser` config given, e.g. `enabled: true`) is now the lightweight operational signals: `control, connection, system, captcha`. The CDP categories (`console, network, page, interaction`) and `screenshot` are dropped from the default and must be opted into.

### Why

"Just turn it on" should not spin up the CDP collector or unleash the high-volume streams. With this default, the collector only starts when a CDP category is explicitly enabled, and `monitor` (collector health) rides along only then. This is the firehose-safe default.

### Test plan

- [x] Rewrote tests
- [x] Updated telemetry session + collector-failure tests for the non-CDP default
- [x] Full non-e2e suite passes

Opus built and ran the container locally

| Test | Expectation | Result |
|---|---|---|
| **(a)** `PUT {}` (enabled, no categories) | default = `control/connection/system/captcha` on; CDP + screenshot off | ✅ 201; exactly those 4 on |
| **(b)** `PUT {console:true}` | opt-in: only `console` captured | ✅ 200; config shows `["console"]` only |
| **publish filter** under console-only | `console_log`→admitted, `page_navigation`→filtered | ✅ 200 / 204 (server-authoritative category + opt-in filter working) |
| **(c)** unknown category `future_cat` | ignored, not rejected | ✅ 200 (forward-compat holds) |
| **(d)** all-off | clears config | ✅ 200, then GET → 404 |
| **firehose safety** | default does NOT start CDP collector; CDP category does | ✅ no `cdpmonitor: started` on default PUT; started only on `console` opt-in |

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 71d3c047998c8834084b97ce8d6e8e8a004bbd81. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


archandatta (approved):
lgtm!

PR #278, opened 2026-06-09
Add replay=all to telemetry stream for replay-from-oldest-retained

archandatta:
## Summary

Adds a `replay=all` query parameter to `GET /telemetry/stream`. When present, the SSE stream starts from the **oldest retained** ring-buffer event (`NewReader(0)`) instead of from-now. This is the one starting position the existing `Last-Event-ID` protocol could not express — `Last-Event-ID: 0` and absent both already mean "from current," and resume is strictly exclusive, so no header value yields the genuine first retained event.

### Behavior

| `Last-Event-ID` | `replay` | result |
|---|---|---|
| absent/empty | `all` | from oldest retained (new) |
| absent/empty | absent/other | from-now (unchanged) |
| `0` | `all` | from-now — `0` keeps its from-current contract |
| `N` (≥1) | `all` | resume after `N` (Last-Event-ID wins) |

`Last-Event-ID` wins over `replay=all` so SSE auto-reconnect (which keeps `?replay=all` on the URL and re-attaches `Last-Event-ID` from the last frame) resumes from the last seen event rather than re-replaying the whole buffer.

The endpoint stays 200-only. `enum: [all]` is documented but not runtime-enforced (no request-validation middleware on the router), so any non-`all` value binds and falls through to from-now rather than returning 400 — a deliberate choice to keep the surface lax and the endpoint additive.

### Changes

- `server/openapi.yaml` — add `replay` query param + `Last-Event-ID` precedence note.
- `server/lib/oapi/oapi.go` — regenerated via `make oapi-generate` (the large diff is the embedded base64 spec blob; ~44 lines are real generated code for the new param).
- `server/cmd/api/api/events.go` — resolve `afterSeq` with Last-Event-ID-wins precedence; `replay=all` → `NewReader(0)`.
- `server/cmd/api/api/events_test.go` — 6 cases (from-oldest, post-eviction gap, LEI precedence, empty-LEI, regression, `LEI:0 + replay=all` from-now).

The ring buffer is unchanged — `newReader(0)` already starts from the oldest retained envelope; only the HTTP handler newly exposes it.

## Test plan

- [x] `make test-unit` green (race enabled)
- [x] Live-verified against the headless container (`docker build -f images/chromium-headless/image/Dockerfile` + run, real HTTP) — see results below.
- [ ] Downstream (separate PRs): kernel/kernel passthrough, SDK regen, CLI `--replay-all`

### Live verification results

Built and ran the headless image, started a telemetry session, published events, and drove the stream over real HTTP (`curl`). This exercises the generated chi router + query-param binding, which the unit tests (which construct the request object directly) bypass. Observed first-frame `id` per request:

| request | observed | expected | result |
|---|---|---|---|
| `GET /telemetry/stream?replay=all` | `id: 1` | oldest retained (`1`) | pass |
| `GET /telemetry/stream?replay=all` + `Last-Event-ID: 3` | `id: 4` | resume after 3 (LEI wins) | pass |
| `GET /telemetry/stream` (no params) | no frame | from-now, no replay | pass |
| `GET /telemetry/stream?replay=foo` | `HTTP/1.1 200`, no replayed frame | 200 + from-now (no 400) | pass |
| `GET /telemetry/stream?replay=all` + `Last-Event-ID: 0` | no frame | from-now (`0` contract) | pass |
| `GET /telemetry/stream` + `Last-Event-ID: 2` (regression) | `id: 3` | resume after 2, unchanged | pass |
| `replay=all` after overfilling the ring (>cap) | first `id` > 1 | eviction gap signal | pass |

Notes:
- The served `/spec.yaml` from the running container included the new `replay` param (confirms the regenerated embedded spec shipped).
- `replay=foo` returning **200** (not 400) and behaving as from-now confirms the documented lenient handling: the router binds any string and the handler falls through when the value is not `all`.
- The overfill case was driven via an in-process `httptest` server against the same generated router; all other cases were driven against the full container on `:10001`.

---

> [!NOTE]
> **Low Risk**
> Additive streaming behavior with explicit precedence rules and broad test coverage; default from-now and Last-Event-ID resume paths are unchanged when replay is not used.
> 
> **Overview**
> Adds **`?replay=all`** on **`GET /telemetry/stream`** so clients can open the SSE feed from the **oldest retained** ring-buffer event (via `NewReader(0)`), not only from the current head or after **`Last-Event-ID`**.
> 
> Stream start position is centralized in **`resolveStartSeq`**: a positive **`Last-Event-ID`** still wins (including over **`replay=all`**, so reconnect does not re-replay history); **`replay=all`** applies when there is no usable header; invalid or non-`all` replay values keep **from-now** behavior (still **200**).
> 
> OpenAPI documents the param and precedence; **`oapi`** is regenerated. Tests cover unit cases for **`resolveStartSeq`** and integration checks for replay, eviction boundaries, and resume regression, with shared **`testRingCapacity`** in API test helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72233d7329e06030c7f9ef292b1c16718a609b5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/cmd/api/api/events.go
```diff
@@ -76,11 +76,21 @@ func (s *ApiService) PublishTelemetryEvent(_ context.Context, req oapi.PublishTe
 func (s *ApiService) StreamTelemetryEvents(ctx context.Context, req oapi.StreamTelemetryEventsRequestObject) (oapi.StreamTelemetryEventsResponseObject, error) {
 	// Default to the current seq so fresh connections only see new events.
 	// Seqs are process-monotonic; a Last-Event-ID from any prior session resumes correctly.
+	// Last-Event-ID wins over replay=all so SSE auto-reconnect resumes from the last
+	// seen event rather than re-replaying history. Any non-empty Last-Event-ID takes
+	// this branch and suppresses replay=all; an unparseable or non-positive value
+	// (including 0) resolves to from-now.
 	afterSeq := s.eventStream.Seq()
```
Sayan-: nit (optional): consider extracting the start-position decision into a pure helper, e.g. `resolveStartSeq(lastEventID *string, replay *oapi.ReplayParam, current uint64) uint64`. it separates the precedence policy (LEI-wins / `0`→from-now / `replay=all`→oldest) from the SSE transport, and lets the 6 precedence cases become a trivial input→seq table test instead of each spinning up a session + pipe + goroutine to assert one number. also avoids the wasted `Seq()` mutex acquire on the `replay=all` path. not blocking.

Sayan- (approved):
lgtm. correctness, concurrency, and consistency all check out against the ring buffer semantics, and the precedence table is well covered. one optional structural nit inline.

PR #285, opened 2026-06-11
Add GET /telemetry/events to read archived telemetry from S2

archandatta:
## Summary

Adds a paginated read path for browser telemetry archived in S2, so events stay available after they age out of the live SSE ring buffer (`GET /telemetry/stream`).

- **`GET /telemetry/events`** — reads archived telemetry envelopes for the browser in ascending sequence order. Returns a top-level JSON array of envelopes plus `X-Has-More` / `X-Next-Offset` pagination headers. Empty result serializes as `[]`, never `null`.
- **Cursor pagination over S2** — `offset` is an S2 sequence cursor (pass the previous response's `X-Next-Offset`); `since`/`until` bound the time window; `category` filters; `limit` is the page size (default 100, max 1000). `has_more` is derived from the S2 read's next-position vs the stream tail, so it is correct whether a page stops on the count, byte, or time bound. Negative `since`/`until`/`offset` are clamped to 0 (`minimum: 0` in the spec; no request-validation middleware is mounted).
- **`events.S2Reader`** — one-shot, bounded S2 reader injected into `ApiService`; builds its S2 client once and reuses it across pages.
- **Configurable batcher** — `S2_BATCHER_LINGER` / `S2_BATCHER_MAX_RECORDS` env vars wire into the existing `S2Config` (defaults unchanged: 100ms / 50).
- **Category schema unified** — the telemetry category enum is promoted to a shared `TelemetryEventCategory` schema; both the read (`TelemetryEvent`) and write (`PublishEventRequest`) sides reference it, so adding a category can't drift the two apart.
- **Fix: telemetry read endpoints no longer self-emit `api_call` events.** The per-request `api_call` middleware was firing for `ReadTelemetryEvents` / `StreamTelemetryEvents`, appending to the very stream being read — a feedback loop that, at page size 1, prevented a paginated read from ever catching the tail. Reads are now side-effect-free.

Envelope/event schemas are unchanged. The response is encoded directly from `events.Envelope` (matching the SSE stream and publish endpoints), so all telemetry endpoints emit an identical envelope shape.

Consumers paginate via `X-Next-Offset` (an S2 stream position) — not the envelope's `seq` field; the param/header docs call this out.

This endpoint is harmless until a consumer calls it, so it can ship independently. A follow-up on the control-plane API will mirror it and wire Stainless `offset_pagination` so the SDK auto-paginates.

### Note on category-filtered pagination

The category filter is applied after the cursor-bounded S2 read (S2 has no server-side filtering), so a filtered page can come back empty while `X-Has-More` is true. `has_more` is authoritative — a client pages until it is false, not until a page is empty. Verified end-to-end (see below).

## Test plan

- [x] Unit: handler (S2-disabled empty path), `buildReadOptions` (offset-over-since precedence, negative-bound clamping, `Count` mapping), `pageSize` clamping, `filterByCategory`, and a middleware regression asserting the telemetry read endpoints emit no `api_call`.
- [x] e2e (`server/e2e`, real S2 via `KI_E2E_BACKEND=docker`): round-trip, category filter, empty-window `[]`, multi-page cursor walk (ascending, no gaps/dupes), category-filtered pagination across empty intermediate pages, and a read-is-side-effect-free assertion. Auto-skips without `S2_BASIN`/`S2_ACCESS_TOKEN`/`S2_STREAM`.
- [x] `go build`, `go vet`, package tests green; `make oapi-generate` output committed.

### Live runs

Ran against a real S2 stream (freshly built headless image, `KI_E2E_BACKEND=docker`):

```
--- PASS: TestReadTelemetryEvents
--- PASS: TestReadTelemetryEventsPagination
--- PASS: TestReadTelemetryEventsFilteredPagination
ok  github.com/kernel/kernel-images/server/e2e
```

Also exercised by hand over HTTP at scale — published ~1000 events and paged the full stream at `limit=50`:

```
pages walked: 19
total_records=950  min_seq=1  max_seq=950  duplicates=0  gaps=0
RESULT: PASS — full stream 1..950 returned exactly once, contiguous, in order
```

`limit=1` terminates cleanly (the feedback-loop fix), and a windowed (`until`) read terminates even while the stream is actively appended (one benign empty boundary page).

---

> [!NOTE]
> **Medium Risk**
> New external read API and S2 pagination semantics affect telemetry consumers; the middleware skip fixes read side-effects but changes observable `api_call` volume on telemetry GETs.
> 
> **Overview**
> Adds **paginated reads of archived telemetry** from durable S2 storage so clients can recover events after the live SSE ring buffer evicts them.
> 
> **`GET /telemetry/events`** returns a JSON array of envelopes (never `null`) with **`X-Has-More` / `X-Next-Offset`** cursor pagination. Query params support `offset` (S2 seq cursor), `since`/`until` (RFC-3339 or durations like `5m`), `limit` (default 20, max 100), and post-read **`category`** filtering. When S2 is not configured, the handler returns an empty `[]`.
> 
> **`events.S2Reader`** performs bounded one-shot S2 reads with tail-aware `has_more` via `nextCursor`; **`ApiService`** takes an optional reader and **`main`** wires `NewS2Reader` alongside the existing writer. S2 append batching is tunable via **`S2_BATCHER_LINGER`** / **`S2_BATCHER_MAX_RECORDS`**, and **`Config.S2Enabled()`** centralizes the enable check.
> 
> **Telemetry read endpoints no longer emit `api_call` events** (`ReadTelemetryEvents`, `StreamTelemetryEvents`), fixing a feedback loop that could prevent pagination from reaching the stream tail.
> 
> OpenAPI and generated clients gain the new operation; publish/read paths share **`TelemetryEventCategory`** instead of a separate publish-only enum. E2E tests exercise real S2 when creds are present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26ab0a1e868cd81c634e6796b4f61562c22d6380. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (commented):
reviewed — solid PR. core pagination mechanics are correct (i traced the S2 cursor/`has_more` termination through the SDK) and match the main API's `offset_pagination` convention; unit + e2e coverage is thorough. main things are spec-surface wording and a couple of consistency items before this gets relayed to the public API. no correctness bugs.

### API surface — keep implementation out of the spec (this gets relayed to the public API)
- `server/openapi.yaml:1383` (summary) — "Read archived telemetry events from durable storage" is implementation-heavy → something like "Read telemetry events for a browser session"
- `server/openapi.yaml:1385-1390` (description) — drop the S2 / "age out of the live SSE ring buffer" / "durable storage is not configured" detail → e.g. "Reads telemetry event data for the browser session."
- `server/openapi.yaml:1398` (offset param) — "This is a stream position" leaks implementation; reword as an opaque pagination cursor. keep the useful guidance (pass back `X-Next-Offset`; don't derive it from the response body / it's not the envelope `seq`).

### Pagination consistency (vs `kernel/kernel` `packages/api/openapi.yaml`)
- core mechanics MATCH: header-based offset pagination, top-level array body, `offset`/`limit` params, `X-Has-More`/`X-Next-Offset` headers — i.e. the Stainless `offset_pagination` model. the planned SDK wiring should work as-is. 
- `server/openapi.yaml:1428` (limit max) — max is **1000**; every main-API list endpoint caps `limit` at **100** (default 20). align to 100, or call out the higher cap as a deliberate telemetry exception.
- `server/openapi.yaml:1406` / `:1414` (since/until) — typed as int64 unix **ms**. the main API's `since` is a **string** (RFC-3339 or a duration like `5m`, e.g. `packages/api/openapi.yaml:8346`) and there's **no `until`** anywhere. prefer RFC-3339/duration for `since` to stay consistent for the public relay, and treat `until` as a documented telemetry-specific addition.
- `offset` taking precedence over `since` is new behavior (no main-API endpoint combines a cursor with a time filter) — worth an explicit doc note.

### Handler
- `server/cmd/api/api/events.go:142` (`defaultReadWindow`) — only sets the *start* (now−5m); the end is bounded by `count`, not a window. reword the name/comment to "default start" to avoid implying an end bound.
- `server/cmd/api/api/events.go:190-195` (`buildReadOptions`) — `until`-only requests (no `since`/`offset`) still anchor the start at now−5m via the default branch, so `until` < now−5m silently yields an empty page. consistent with the documented `since` default, but a mild footgun — confirm intended, or special-case until-only to start from the beginning.

### Test coverage
- the core `S2Reader.Read` + `has_more`/cursor logic (the subtlest code here) is exercised **only** by the e2e tests, which auto-skip without S2 creds + docker. worth confirming CI runs them with creds; otherwise an interface seam over the S2 read session for a mocked unit test would give fast-feedback coverage of the cursor math on every run.
- windowed (`until`) pagination termination on an actively-appended stream is hand-verified only (the "one benign empty boundary page" in the description), not in a committed test. my read of the SDK agrees it terminates (a zero-record page → `NextReadPosition()==nil` → `has_more=false`), but consider an e2e guard.

### Nits (optional)
- duplicate "is S2 configured?" predicate — the explicit three-field `if config.S2Basin != "" && …` in `main.go` for the writer vs `NewS2Reader`'s internal copy of the same check; a small `config.S2Enabled()` helper would keep them from drifting.
- config defaults (`100ms`/`50`) live in both the config struct tags and the `s2storage.go` zero-value fallbacks (already acknowledged in a comment) — just duplication to be aware of.
- `S2Reader.Read` unbounded-read guard (`Count==nil && Until==nil → Until=now`) is dead from the only caller (the handler always sets `Count`) and untested — fine as defensive library code, just noting it's unreachable from this path.

rgarcia (approved):

rgarcia (changes requested):
actually after chatting with @Sayan- I think we should make this endpoint control-plane only so we don't have any ambiguity about whether an SDK client should go direct to VM or to the control plane. So let's close this and port the logic into the control plane instead

archandatta:
> actually after chatting with @Sayan- I think we should make this endpoint control-plane only so we don't have any ambiguity about whether an SDK client should go direct to VM or to the control plane. So let's close this and port the logic into the control plane instead

Ah gotcha, yep I will have a PR ready today in the other repo for this!

PR #286, opened 2026-06-11
fix(e2e,chromedriverproxy): WebSocket/CDP scheme over a TLS-terminating ingress

rgarcia:
ChromeDriver BiDi and CDP over a TLS-terminating ingress (the hypeman :9224/:9222 ingresses; serve https/wss) were broken because the WebSocket/CDP URLs used the wrong scheme. Docker is plaintext (http/ws) so it never showed there. Validated in kernel-images-private #237: the full e2e suite (incl. all 5 TestBidi* and TestCDPProxyReconnect) now passes against the staging hypeman host over Tailscale.

## Fixes
1. `server/e2e/container.go` — `ChromeDriverAddr` only stripped `http://` (returned `https://host:9224` with scheme attached); `ChromeDriverWSURL` hardcoded `ws://` (→ malformed `ws://https://host:9224/session`). Now strip any scheme and pick `ws`/`wss` from the ChromeDriver URL's transport.
2. `server/lib/chromedriverproxy/proxy.go` — `rewriteWebSocketURL` kept chromedriver's `ws://` scheme, so the `webSocketUrl` from `POST /session` was unreachable through the TLS ingress. New `clientWSScheme(r)` returns `wss` when `X-Forwarded-Proto: https` (Caddy ingress) or `r.TLS != nil`; docker keeps `ws`. + regression test.
3. `server/e2e/e2e_cdp_reconnect_test.go` — `fetchBrowserWebSocketURL` hardcoded `http://` for the CDP `/json/version` probe (fails against the TLS `:9222` ingress). Derive the scheme from the CDP URL's transport (wss → https).

---

> [!NOTE]
> **Medium Risk**
> Touches ChromeDriver session response rewriting and e2e URL construction; incorrect scheme logic would break remote BiDi/CDP but changes are narrowly scoped with new regression tests.
> 
> **Overview**
> Fixes **BiDi and CDP** when traffic goes through the hypeman TLS ingress (`https`/`wss` on :9224/:9222). Plain Docker `http`/`ws` behavior is unchanged.
> 
> **ChromeDriver proxy** — On `POST /session`, `rewriteWebSocketURL` now sets `capabilities.webSocketUrl` to **`wss://`** when the client reached the proxy via TLS (`X-Forwarded-Proto: https` / `r.TLS`), instead of leaving ChromeDriver’s internal **`ws://`** URL. New `clientWSScheme` handles comma-separated forwarded proto values. Regression test added.
> 
> **E2E helpers** — `ChromeDriverAddr` strips any URL scheme (not only `http://`); `ChromeDriverWSURL` picks **`ws` vs `wss`** from the ChromeDriver base URL. `fetchBrowserWebSocketURL` uses **`https`** for `/json/version` when `CDPURL()` is `wss://`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae8abde84068b78228a31bb49e41c991afec04f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/e2e/container.go
+++ b/server/e2e/container.go
@@ -76,16 +76,27 @@ func (c *TestContainer) ChromeDriverURL() string {
 }
 
 // ChromeDriverAddr returns the host:port for the instance's ChromeDriver proxy,
-// derived from ChromeDriverURL (without scheme). Useful for substring assertions
-// on proxy-rewritten URLs.
+// derived from ChromeDriverURL (scheme stripped). Useful for substring
+// assertions on proxy-rewritten URLs. Handles both http:// (docker) and
+// https:// (hypeman ingress) ChromeDriver URLs.
 func (c *TestContainer) ChromeDriverAddr() string {
-	return strings.TrimPrefix(c.backend.ChromeDriverURL(), "http://")
+	u := c.backend.ChromeDriverURL()
+	if i := strings.Index(u, "://"); i >= 0 {
+		return u[i+3:]
+	}
+	return u
 }
 
-// ChromeDriverWSURL returns the WebSocket URL (ws://host:port/path) for the
-// instance's ChromeDriver proxy. path should include a leading slash.
+// ChromeDriverWSURL returns the WebSocket URL for the instance's ChromeDriver
+// proxy. The scheme matches the ChromeDriver endpoint's transport: wss:// when
+// it's served over TLS (the hypeman ingress on :9224), ws:// otherwise (docker).
+// path should include a leading slash.
 func (c *TestContainer) ChromeDriverWSURL(path string) string {
-	return "ws://" + c.ChromeDriverAddr() + path
+	scheme := "ws"
+	if strings.HasPrefix(c.backend.ChromeDriverURL(), "https://") {
+		scheme = "wss"
+	}
+	return scheme + "://" + c.ChromeDriverAddr() + path
 }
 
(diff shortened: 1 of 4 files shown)


tnsardesai (approved):
reviewed — clean, well-scoped fix with a focused regression test, and the docker plaintext path stays untouched via the `""` scheme sentinel. lgtm. one thing worth a look:

### Questions
- `server/lib/chromedriverproxy/proxy.go:192` — `X-Forwarded-Proto` can be comma-separated (`https, http`) when a request traverses multiple proxies. `r.Header.Get` returns the raw first header value, so `EqualFold(proto, "https")` won't match `"https, http"` and falls through to `"ws"` — silently reintroducing the bug this PR fixes. worth confirming the Caddy ingress only ever sets a single value; if there's any doubt, splitting on `,` and taking the first token is cheap insurance.

rgarcia:
Thanks @tnsardesai — good catch. Fixed in `ae8abde`: `clientWSScheme` now splits `X-Forwarded-Proto` on `,` and uses the first (client-facing) token before comparing, so a multi-proxy `"https, http"` value still yields `wss`. Added a `"https, http"` case to `TestHandler_PostSession_WSSchemeFromForwardedProto`. (Mirrored the same fix into the private hypeman-e2e PR.)

PR #13, opened 2025-06-23
WebRTC OSS launch

juecd:
# Checklist

- [ ] A link to a related issue in our repository
- [x] A description of the changes proposed in the pull request.
- [ ] @mentions of the person or team responsible for reviewing proposed changes.

Updates README
Renames folder structure to `images` -> `/chromium-headless` and `/chromium-headful`
Updates the WebRTC favicon

PR #288, opened 2026-06-12
Fix kmsg timestamps

Sayan-:
### Overview

`system_oom_kill` events were stamped with the kmsg envelope timestamp, which is `CLOCK_MONOTONIC`-derived. That clock freezes while a unikraft VM is suspended (scale-to-zero), so on any VM that had standbyed the OOM timestamps skewed backward by the accumulated suspended duration — sometimes by hours.

- `sysmon` kmsg reader now stamps each record with wall-clock read time (`time.Now()`) instead of the envelope timestamp. We only ever read live records (the source seeks to end first), so read time is an accurate event time.
- Documented the invariant on `events.Event.Ts` and added a gotcha (#13) in `AGENTS.md`: all event timestamps must be wall-clock at emit/observe, never a monotonic/source clock.

### Test
- Added a linux-gated unit test asserting `kmsgparserSource` stamps observation time, not the envelope's monotonic time

---

> [!NOTE]
> **Low Risk**
> Narrow telemetry timestamp fix in sysmon with documentation and a focused unit test; no auth, API contract, or OOM parsing logic changes.
> 
> **Overview**
> Fixes **`system_oom_kill`** telemetry showing times far in the past on scale-to-zero VMs by no longer using the kmsg envelope’s monotonic-derived clock (which freezes while the VM is suspended).
> 
> **`kmsgparserSource`** now sets each `KmsgMessage.Timestamp` to **`time.Now()`** at read time instead of the parser’s envelope timestamp, so OOM **`TimeOfDeath`** and downstream **`events.Event.Ts`** reflect when the record was observed. Comments on **`events.Event.Ts`**, **`KmsgMessage`**, and **`OomInstance.TimeOfDeath`** document the wall-clock-at-emit/observe rule; **AGENTS.md** adds gotcha **#13** for all in-process producers.
> 
> Adds a **linux-gated** unit test with a fake kmsg parser to assert observation time is stamped, not the envelope time.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c140ce46fc3e7299951f238c80081c9fc4768a44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (approved):
approving — clean, low-risk fix. traced the chain end to end: `time.Now()` (kmsg_linux.go:48) → `KmsgMessage.Timestamp` → `feed()` (sysmon.go:116) → `OomInstance.TimeOfDeath` (kmsg.go:190) → `Event.Ts` (sysmon.go:178), so the wall-clock stamp genuinely reaches the published OOM event. the `SeekEnd()` precondition the comment relies on holds (kmsg_linux.go:23), and the new test passes locally.

a few doc/test nits, none blocking:

### nits
- `AGENTS.md:52` (gotcha #13) — lists the supervisord shim as an in-process producer that "must stamp `time.Now()` themselves", but it's actually an HTTP producer: `cmd/supervisord-shim` POSTs a `PublishEventRequest` with no `ts` (main.go:236), so it gets stamped at ingest by `TelemetrySession.publishLocked` (telemetry.go:76-77) — i.e. the doc's *other* clause. consider moving it to the "HTTP-published … stamped by the API handler" side, or swapping in a genuine in-process example (e.g. cdpmonitor, which stamps `time.Now().UnixMicro()` itself).
- `AGENTS.md:52` — since `publishLocked` already defaults a zero `Ts` to wall-clock, the precise hazard #13 guards against is a producer setting `Ts` to a *non-zero, non-wall-clock* value (exactly the envelope bug here). wording it that way would make the "why" land harder.
- `kmsg_linux_test.go:52` — the `!Equal(envelopeTime)` assert is redundant with the `[before, after]` bracket (`before` is decades after the envelope time, so `!Before(before)` already excludes it). harmless and documents intent — optional to drop.

nice touch documenting the invariant on `Event.Ts` itself (event.go) so the next producer author sees it at the field.

archandatta (approved):
lgtm!

PR #289, opened 2026-06-15
Add named markers to replay recordings as MP4 chapters

rgarcia:
## What

Adds named, timestamped markers to screen recordings. A marker records a point in time during an active recording; at finalize the markers are injected into the output MP4 as **chapter markers**, so users can scrub/crop to them in post (ffprobe `-show_chapters`, QuickTime, VLC, and most NLEs all read MP4 chapters).

The feature is purely additive and backwards-compatible.

## Why

Recordings are currently opaque blobs — there's no way to flag "the interesting thing happened here" while a session runs. MP4 chapters are the standard, tool-agnostic way to annotate timeline positions, so a marker dropped during recording survives into the downloaded file with no extra sidecar.

## API

New endpoint `POST /recording/mark`:

- Body: `{ "name": string (required), "id"?: string }`
- `201` `{ "name": string, "offsetMs": integer }` on success
- `409` when no recording is in progress
- `400` for an empty/oversized name or missing body

`offsetMs` is documented as **provisional** (measured against the recording start time at mark time); the authoritative offset is the chapter start computed at finalize.

## How

- `Mark(name)` on the recorder appends a marker under the existing mutex and is safe under concurrent calls.
- At finalize, when markers exist, an ffmetadata file is built (integer-millisecond `TIMEBASE=1/1000` chapters, special characters escaped) and passed as a second input with `-map 0 -map_chapters 1`. A sentinel chapter is prepended at `START=0` because ffmpeg forces the first chapter to start at 0 — without it the first real marker's timestamp would be clamped.
- With **no markers**, the remux command is byte-identical to before. Any marker/metadata failure falls back to the plain remux, so a recording is never failed or corrupted by this path.
- The OpenAPI spec is the source of truth; `lib/oapi/oapi.go` was regenerated via `make oapi-generate` (openapi-down-convert + oapi-codegen), not hand-edited.

## Tests

- Unit tests for the chapter-metadata builder: sentinel at `START=0`, integer `START`/`END` tiling, negative-offset markers dropped, ascending ordering, special-character escaping, and a nonzero origin-shift parameter.
- Unit tests for `Mark()`: 409 path when not recording, trim/validation, and concurrency (run under `-race`).
- A backwards-compat test locking the no-marker remux args to the pre-feature command and asserting `-map_chapters` is injected only when markers are present.
- An integration test (real ffmpeg, lavfi `testsrc`, no display needed) that records a synthetic clip, marks at known offsets, finalizes, and asserts via `ffprobe -show_chapters` that chapters land within ~1.5 frame intervals.

`go vet ./...`, `go build ./...`, and the full non-e2e suite (`go test -race`) pass locally.

---

> [!NOTE]
> **Medium Risk**
> Touches finalize/remux and ffmpeg invocation on every stopped recording when markers exist; failures are mitigated by fallback remux, but timing/anchor logic could still misplace chapters without corrupting the file.
> 
> **Overview**
> Adds **`POST /recording/mark`** so clients can drop named points on an active recording session (optional `id`, same as other `/recording/*` endpoints). Success returns **201** with the trimmed name and provisional **`offset_ms`**; errors map to **400** (missing body / invalid name), **404** (unknown session), and **409** (not actively recording).
> 
> The **`Recorder`** / **`FFmpegRecorder`** stack gains **`Mark(name)`**, which stores markers under the existing mutex. On **finalize**, markers become **MP4 chapters**: an **ffmetadata** sidecar is built (sentinel chapter at 0 when needed, sorted chapters, escaped titles), remuxed with **`-map_chapters`**, and **chapter remux failures fall back** to the previous plain remux so recordings are not lost. **No markers** keeps the remux argv **byte-identical** to before. **`captureAnchor`** (first output bytes) and stale output removal at start align chapter times with the media timeline.
> 
> OpenAPI and **`lib/oapi`** are regenerated; recording-session **`id`** field descriptions are clarified across start/stop/mark/download/delete.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4135b06aee02cce593e2aad4b12c8e293685b91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sjmiller609 (approved):

sjmiller609 (approved):
I think it needs separate handling in GPU mode btw

PR #296, opened 2026-06-24
Close hijacked WebSockets with 1001 on shutdown

Sayan-:
### Summary
- `http.Server.Shutdown` does not close or wait for hijacked connections. The proxy's hijacked WebSocket flows (CDP proxy, WebDriver/BiDi, ChromeDriver, PTY attach) were torn down abruptly on shutdown, so clients observed a `1006 abnormal closure` instead of a clean `1001 Going Away`.
- Add a small `wsdrain.Registry` that tracks live client connections. 
- Because the first WebSocket close wins, each proxy's own normal-closure teardown no longer overrides the code the client sees.

### Testing

- [x] CI + added coverage
- [x] Ran locally + manually connect Playwright over CDP, stop the server, confirm client close code is 1001

---

> [!NOTE]
> **Medium Risk**
> Changes the shutdown path and every hijacked WebSocket entry point, but behavior is additive (tracking + explicit close) with tests and nil-safe wiring; mis-ordering shutdown could still leave brief races with in-flight sessions.
> 
> **Overview**
> **Hijacked WebSocket sessions** (CDP DevTools proxy, ChromeDriver/BiDi, PTY attach) were not drained by `http.Server.Shutdown`, so clients often saw **1006 abnormal closure** instead of a deliberate server shutdown.
> 
> This PR adds **`wsdrain.Registry`**: each accepted client connection is **tracked** after upgrade and **untracked** on teardown. On graceful shutdown, **`CloseAll(1001 Going Away, …)`** runs in parallel with HTTP shutdown so clients get a clean close before proxy goroutines send their usual normal closure.
> 
> The same registry is wired through **`wsproxy.Proxy`**, the DevTools proxy (direct accept path), **PTY attach**, and **chromedriver** WebSocket proxy. A **nil registry** is a no-op so tests stay simple. New unit tests cover registry behavior and DevTools proxy **1001** visibility when `CloseAll` wins the first-close race.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0eb856d826200b083ca9f8292714eb2bc25f62ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/cmd/api/api/process.go
+++ b/server/cmd/api/api/process.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kernel/kernel-images/server/lib/logger"
 	oapi "github.com/kernel/kernel-images/server/lib/oapi"
 	"github.com/kernel/kernel-images/server/lib/ptyio"
+	"github.com/kernel/kernel-images/server/lib/wsdrain"
 	openapi_types "github.com/oapi-codegen/runtime/types"
 )
 
@@ -659,7 +660,7 @@ func writeJSON(w http.ResponseWriter, status int, body string) {
 //   - Server sends TextMessage with JSON for events (e.g., exit code)
 //
 // This endpoint is intentionally not defined in OpenAPI.
-func (s *ApiService) HandleProcessAttachWS(w http.ResponseWriter, r *http.Request, id string) {
+func (s *ApiService) HandleProcessAttachWS(w http.ResponseWriter, r *http.Request, id string, reg *wsdrain.Registry) {
 	ctx := r.Context()
 	log := logger.FromContext(ctx)
 
@@ -701,6 +702,9 @@ func (s *ApiService) HandleProcessAttachWS(w http.ResponseWriter, r *http.Reques
 		return
 	}
 	defer wsConn.CloseNow()
+
+	untrack := reg.Track(wsConn)
+	defer untrack()
 
 	// Set a generous read limit for PTY data
 	wsConn.SetReadLimit(1024 * 1024) // 1MB
--- a/server/cmd/api/main.go
+++ b/server/cmd/api/main.go
@@ -47,11 +47,6 @@ func main() {
 	}
 	slogger.Info("server configuration", "config", config)
 
-	// Tracks live hijacked WebSocket connections (CDP, WebDriver/BiDi,
-	// ChromeDriver, PTY attach) so they can be closed with a clean Going Away
-	// frame on shutdown — http.Server.Shutdown does not touch hijacked conns.
-	wsRegistry := wsdrain.New()
-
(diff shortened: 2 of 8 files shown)


hiroTamada (commented):
reviewed incrementally — the core design is sound (nil-safe registry, correct track/untrack ordering at every call site, first-close-wins proven by a real integration test). one thing i'd want bounded or explicitly acknowledged before this goes in, plus a smaller race you already know about.

### shutdown latency — worth bounding before merge
- `server/lib/wsdrain/wsdrain.go:64-66` — `CloseAll` closes conns sequentially, and coder/websocket's `Close` runs a full handshake: it writes the close frame then waits up to 5s for the peer to echo, on its own internal timeout (it takes no context, so `shutdownCtx` doesn't bound it). today `http.Server.Shutdown` returns *without* waiting on hijacked WS conns, so this goroutine adds up-to-N×5s of blocking that wasn't on the shutdown path before — against the single 10s budget in `main.go`, with `g.Wait()` blocking on it. usually it's a single CDP client that echoes instantly, but 2-3 stuck/half-open clients (CDP + chromedriver + PTY can all be live) could push shutdown past the budget. consider closing each conn in its own goroutine + `sync.WaitGroup` (optionally a short cap) so total is ~5s regardless of count.

### known race (already noted in the description)
- `server/cmd/api/main.go:294-303` — `CloseAll` runs concurrently with the server drains + `upstreamMgr.Stop()`, so a proxy's own `NormalClosure` teardown can win the first-close race and the client sees 1000 (or 1006). best-effort by design, and coupled to the above: a bounded-latency `CloseAll` would let you sequence it *before* the drains for a deterministic 1001 without risking the budget.

### nits
- `server/lib/wsdrain/wsdrain_test.go` — no concurrency/`-race` test, though the registry exists specifically to be hit concurrently on shutdown. optional.
- `server/cmd/api/api/process.go:704` — PTY attach tears down with `CloseNow()` (no frame) on normal session end, so non-shutdown closes stay 1006; the shutdown 1001 path is fine since `Close` writes the frame first. pre-existing, just noting.

otherwise lgtm — tests are meaningful and the wiring is clean.

hiroTamada (approved):
approving. the `wsdrain.CloseAll` sequential-close latency i flagged is acknowledged but non-blocking here — VM shutdown is async in the control plane, so the extra up-to-N×5s on the drain path isn't on a user-facing path. design is clean and the tests are meaningful. the 1001-vs-1000 ordering stays best-effort as documented; if a deterministic 1001 is ever wanted, a bounded-latency CloseAll + sequencing it ahead of the drains is the path, but not needed for this.

PR #303, opened 2026-07-02
Add optional in-VM OTLP telemetry export sink

archandatta:
## Summary
- Adds an optional in-VM OTLP/HTTP export sink that converts browser telemetry events into OTLP log records and forwards them to a configured endpoint (an OTLP relay/collector), running alongside the existing S2 sink. Gated on `OTLP_RELAY_ENDPOINT` — no behavior change when unset.
- Converter maps each telemetry envelope to an OTLP log record: event type → `EventName` (also mirrored to a `kernel.event.type` attribute for backends that drop `EventName`), structured JSON body, derived severity, and `kernel.*` attributes. High-value network/console fields are promoted to semantic-convention attributes (`http.response.status_code`, `url.full`, `http.request.method`). Screenshot/monitor categories are excluded from export.
- Sink mirrors `S2StorageWriter`: an independent ring reader feeds records through the OTel log SDK (`otlploghttp` + batch processor). Runs independently of S2 (both can be enabled). Export failures are surfaced via a logging exporter wrapper (the SDK otherwise reports them only through an uninstalled global logger).

## Testing
- Unit tests for the converter (field mapping, severity, structured body, promoted attributes, category exclusion) and an end-to-end sink test that decodes the exported OTLP payload and asserts an excluded category never reaches the receiver.
- Validated locally against `otelcol-contrib` (config under `server/dev/otelcol/`), including a real headless-Chromium CDP run confirming live events convert and export correctly.
- `go build ./...`, `go vet`, and the events test suite pass.

## Notes
- Draft: the export sink is gated off by default and depends on a reachable OTLP relay endpoint in production; not yet wired into deploy config.
- Delivery-substrate benchmarking (relay vs a durable server-side pusher) lives on a separate branch, `hypeship/otlp-substrate-bench`, to keep this PR reviewable in isolation.


archandatta:
Superseded by https://github.com/kernel/kernel-images/pull/308 (same exporter work, re-homed under the KERNEL-1213 branch with squashed history).

PR #307, opened 2026-07-07
Capture interaction events on already-loaded pages; stop dropping keystrokes

archandatta:
## Summary

Interaction telemetry (`interaction_click` / `interaction_key` / `interaction_scroll_settled`) was not captured when a page was already loaded at the moment the monitor attached — e.g. when the `interaction` category is enabled mid-session, or the monitor attaches to a pre-existing tab. Interactions only started flowing after the next navigation. This fixes that and a related keystroke-loss issue.

**Root cause:** `interaction.js` was installed only via `Page.addScriptToEvaluateOnNewDocument`, which runs at the start of the *next* document load and never touches the currently-loaded document. So the click/key/scroll listeners were never attached to a live page until it navigated.

## Changes

- **Current-document injection** (`domains.go`): `injectScript` now also `Runtime.evaluate`s `interaction.js` against the already-loaded document, in addition to registering it for future loads. Idempotent via `window.__kernelEventInjected`. This covers both live-view (X11) and CDP-driven input equally — the discriminator was script presence on the current document, not the input path.
- **Keystroke capture** (`handlers.go`): `interaction_key` is now exempt from the 20/s per-type binding rate limit. Keystrokes can arrive faster than the cap (especially automated/bulk typing), and dropped keys leave recorded input incomplete. Clicks and scrolls remain rate-limited.

## Tests

- Unit: current-document injection sends `Runtime.evaluate` with the tracker source on attach; `interaction_key` is not rate-limited (clicks still are).
- E2E (opt-in, `KERNEL_CDPMONITOR_CHROME_E2E=1`): a real headless Chromium integration test that loads a page, starts the real `Monitor`, then drives a click and rapid typing on the already-loaded page (no re-navigation) and asserts capture.

## Smoke test (real headless Chromium + real `Monitor`, before/after)

| scenario | before | after |
|---|---|---|
| page already loaded when monitor attaches → `window.__kernelEventInjected` | `false` | `true` |
| click on already-loaded page, no re-nav | not captured | `interaction_click` |
| type "hello" (5 keys, fast) | dropped | all 5 `interaction_key` |
| existing `cdpmonitor` unit suite | pass | pass |

The e2e test fails at "interaction.js was not injected" when the `domains.go` fix is reverted and passes with it, confirming it guards the fix.

## Notes

- All changes are in `server/lib/cdpmonitor/`; no generated files touched.
- The e2e test skips by default so normal unit runs never launch a browser.

---

> [!NOTE]
> **Medium Risk**
> Changes live CDP injection and interaction event rate limiting in production telemetry paths; behavior is well tested but affects what downstream consumers see for mid-session pages and fast typing.
> 
> **Overview**
> Fixes missing **interaction** telemetry when the CDP monitor attaches to a tab that is already loaded (e.g. enabling the interaction category mid-session), and reduces **keystroke loss** from binding rate limits.
> 
> **`injectScript`** still registers `interaction.js` for future navigations via `Page.addScriptToEvaluateOnNewDocument`, and now also runs the same script on the **current document** with `Runtime.evaluate` (idempotent via `window.__kernelEventInjected`). Failed eval on non-evaluable pages is warned but does not block the new-document hook.
> 
> **`handleBindingCalled`** applies a **higher per-type rate cap** for `interaction_key` (`bindingKeyMinInterval`, ~1000/s) than clicks/scrolls (`bindingMinInterval`, 20/s), so fast typing is less likely to drop keys while other interaction types stay throttled.
> 
> Coverage adds unit tests for current-document injection and key rate limits, plus an **opt-in** headless Chromium e2e (`KERNEL_CDPMONITOR_CHROME_E2E=1`) that asserts injection, click capture, and full keystroke capture without re-navigation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43d51843c4baded8922075f899a69861130902b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/lib/cdpmonitor/handlers.go
+++ b/server/lib/cdpmonitor/handlers.go
@@ -198,6 +198,21 @@ func (m *Monitor) handleExceptionThrown(ctx context.Context, p cdpRuntimeExcepti
 // a misbehaving page from flooding the event pipeline.
 const bindingMinInterval = 50 * time.Millisecond
 
+// bindingKeyMinInterval caps interaction_key at 1000 events/s. Typing produces
+// keystrokes in bursts faster than other interaction events, and a dropped key
+// leaves the recorded input incomplete, so keys get a higher cap; the cap still
+// bounds a page that synthesises a keydown flood.
+const bindingKeyMinInterval = time.Millisecond
+
+// bindingIntervalFor returns the minimum spacing between accepted binding events
+// of the given type.
+func bindingIntervalFor(eventType string) time.Duration {
+	if eventType == EventInteractionKey {
+		return bindingKeyMinInterval
+	}
+	return bindingMinInterval
+}
+
 // handleBindingCalled processes __kernelEvent binding calls from the page.
 func (m *Monitor) handleBindingCalled(p cdpRuntimeBindingCalledParams, sessionID string) {
 	if p.Name != bindingName {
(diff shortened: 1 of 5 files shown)


server/lib/cdpmonitor/handlers.go
```diff
@@ -272,16 +222,21 @@ func (m *Monitor) handleBindingCalled(p cdpRuntimeBindingCalledParams, sessionID
...
 	// misbehaving page cannot flood the event pipeline with a single event type.
-	now := time.Now()
-	rateKey := sessionID + ":" + header.Type
-	m.bindingRateMu.Lock()
-	last := m.bindingLastSeen[rateKey]
-	if now.Sub(last) < bindingMinInterval {
+	// interaction_key is exempt from the rate limit: keystrokes can arrive faster
+	// than the cap, and a dropped key leaves the recorded input incomplete. Other
+	// event types stay rate-limited unless added to this exemption.
+	if header.Type != EventInteractionKey {
```
Sayan-: instead of unlimited I wonder if splitting and allowing 200-1000 typing events per second is still better than uncapped

Sayan- (changes requested):
qq

Sayan- (approved):
yeet

PR #308, opened 2026-07-07
Add in-VM OTLP telemetry export sink

archandatta:
## Summary
- Adds an optional in-VM OTLP/HTTP export sink that converts browser telemetry events into OTLP log records and forwards them to a configured endpoint, alongside the existing S2 sink. Gated on `BTEL_OTLP_ENDPOINT`; no behavior change when unset.
- Converter maps each telemetry envelope to an OTLP log record: event type to `EventName` (mirrored to a `kernel.event.type` attribute for backends that drop `EventName`), a structured JSON body, derived severity, and `kernel.*` attributes. High-value network/console fields are promoted to semantic-convention attributes (`http.response.status_code`, `url.full`, `http.request.method`). Screenshot and monitor categories are excluded.
- Sink mirrors `S2StorageWriter`: an independent ring reader feeds records through the OTel log SDK (`otlploghttp` + batch processor), running independently of S2.

## Config (`BTEL_OTLP_*`)
- `BTEL_OTLP_ENDPOINT` (host[:port]; presence enables the sink), `BTEL_OTLP_PATH` (default `/v1/logs`), `BTEL_OTLP_INSECURE`, `BTEL_OTLP_SERVICE_NAME` (default `kernel-browser`).
- Platform identity reused for the OTLP Resource and auth header: `KERNEL_INSTANCE_JWT`, `INST_NAME`, `METRO_NAME`.
- The sink is a standard OTLP/HTTP client, so the endpoint can be a local collector, a forwarding relay, or a customer's OTLP backend directly. Pointing at the metro-api relay just means setting `BTEL_OTLP_PATH=/otlp-relay/v1/logs`.

## Review feedback addressed
- Renamed export env vars `OTLP_RELAY_*` to `BTEL_OTLP_*` so they don't collide with the standard `OTEL_*` vars that configure the API server's own telemetry, and dropped the relay-specific naming (dev points at a collector).
- Renamed the platform-identity config fields to `InstanceJWT` / `InstanceName` / `MetroName` (they aren't OTLP-specific).
- Sink start failure is non-fatal: an optional best-effort exporter must not take down the browser, so it logs and continues instead of exiting.
- Wired the OTel SDK's global logger so batch-queue drops under sustained backpressure surface in logs instead of being silently discarded.

## Testing
- Converter unit tests (field mapping, promoted attributes, structured body, severity, truncation, number coercion, JSON fallback, category exclusion), an end-to-end sink test that decodes the exported OTLP payload from a fake receiver and asserts an excluded category never lands, and a logging-exporter test asserting export failures are counted.
- `go build ./...`, `go vet`, and the server unit suite are green.
- Manually validated end-to-end against a local collector and a self-hosted SigNoz with real headless-Chromium CDP runs: page/network/console events land, screenshot/monitor stay excluded, and promoted attributes are queryable. Local recipe:
  ```
  docker run --rm -p 4318:4318 otel/opentelemetry-collector-contrib \
    --set=receivers.otlp.protocols.http.endpoint=0.0.0.0:4318 \
    --set=exporters.debug.verbosity=detailed \
    --set=service.pipelines.logs.receivers=[otlp] \
    --set=service.pipelines.logs.exporters=[debug]
  # then run the server with:
  #   BTEL_OTLP_ENDPOINT=127.0.0.1:4318 BTEL_OTLP_INSECURE=true
  ```

## Scope and follow-ups
- MVP scope: logs, best-effort delivery. Off by default. Producing telemetry for a real browser session additionally requires provisioning to set `BTEL_OTLP_ENDPOINT` on the VM (not yet wired), gated on a resolved destination so we don't generate work with nowhere to land.
- Deferred (tracked separately): spans, durable / at-least-once delivery, and exporter improvements (export enable/disable API, service-crash and OOM-kill events, externalized batch knobs, event reshape for relay routing, a drop metric).

---

> [!NOTE]
> **Medium Risk**
> New async export path and dependency bump affect telemetry delivery and shutdown; misconfiguration could leak JWT in headers or flood backends, but failures are best-effort and off by default.
> 
> **Overview**
> Adds an **optional OTLP/HTTP log export path** for browser telemetry, parallel to the existing S2 sink. When `BTEL_OTLP_ENDPOINT` is set, the API starts an `OTLPStorageWriter` that reads the same event ring, converts envelopes to OTLP log records, and batches them via the OTel log SDK (`otlploghttp`).
> 
> **Configuration** uses `BTEL_OTLP_*` env vars (to avoid clashing with `OTEL_*` for the API’s own telemetry), plus existing platform identity (`KERNEL_INSTANCE_JWT`, `INST_NAME`, `METRO_NAME`) for the OTLP resource and optional `Authorization` header.
> 
> **Converter behavior:** maps event type to `EventName` and mirrors it to `kernel.event.type`; structured JSON body; coarse severity from type strings; `kernel.*` and promoted HTTP/console attributes. **Screenshot** and **monitor** categories are not exported.
> 
> **Operational choices:** OTLP start failure is **non-fatal** (logs and continues, unlike S2). On successful start, the OTel SDK global logger is wired so batch-queue drops are visible. Shutdown drains the writer like S2. Export requests are **chunked** (~4 MiB) to stay under typical HTTP body limits.
> 
> Dependencies add OpenTelemetry log SDK and OTLP HTTP exporter; unit and integration tests cover conversion, chunking, export failures, and end-to-end export to a fake receiver.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69804f68a8ed32f13140bb5fd0423608c68e7a17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (commented):
reviewed — overall looks good. the converter/sink split is clean, mirroring the S2 writer lifecycle was the right call, and the e2e test decoding real OTLP protobuf off the wire is solid. two naming findings:

- `server/cmd/config/config.go:54-56` — env var naming: consider prefixing with something browser-telemetry-specific (e.g. `BTEL_*`) so these aren't confused with standard `OTEL_*` env vars that would control kernel-images-api's own telemetry; also `RELAY` feels overly specified since the endpoint isn't necessarily a relay (a collector in dev)
- `server/cmd/config/config.go:60-62` — nit: field names should be `InstanceJWT` / `InstanceName` / `MetroName` — they're platform identity envs, not OTLP-specific

server/cmd/config/config.go
```diff
@@ -47,6 +47,19 @@ type Config struct {
 	S2Basin       string `envconfig:"S2_BASIN"        default:""`
 	S2AccessToken string `envconfig:"S2_ACCESS_TOKEN" default:""`
 	S2Stream      string `envconfig:"S2_STREAM"       default:""`
+
+	// OTLP telemetry export. OTLPEndpoint (host[:port]) must be set to enable
+	// the OTLP sink, which forwards telemetry events to the metro-api relay (or
```
Sayan-: should drop refs to metro-api

server/lib/events/otlpconvert.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return out
+}
+
+// otlpSeverity maps an event type to an OTLP severity. It is a deliberately
+// coarse, best-effort mapping keyed off the producer's event-type string
+// conventions (the literal "console_error" and the "_failed" suffix), not the
+// generated event-type constants: those live in the producer package
+// (cdpmonitor), which this package cannot import without a cycle. If those type
+// names change, update this mapping in lockstep.
```
Sayan-, line 125: this seems fine as a starting point. I think it would be helpful to mux on service crash and oom kill events as well

server/lib/events/otlpstorage.go
```diff
@@ -0,0 +1,197 @@
...
+
+// otlpStorage implements Storage by converting envelopes to OTLP log records
+// and emitting them through the OTel log SDK, which batches and exports to the
+// configured endpoint.
+type otlpStorage struct {
+	provider *sdklog.LoggerProvider
+	logger   log.Logger
+}
+
+func newOTLPStorage(ctx context.Context, cfg OTLPConfig, log *slog.Logger) (*otlpStorage, error) {
```
Sayan-, line 139: I think this is fine as a starting point. I'd like for us to consider how much of these we want to control externally

server/cmd/api/main.go
```diff
@@ -130,6 +125,29 @@ func main() {
...
+		otlpWriter = events.NewOTLPStorageWriter(eventStream, events.OTLPConfig{
+			Endpoint:     config.OTLPEndpoint,
+			URLPath:      config.OTLPPath,
+			Insecure:     config.OTLPInsecure,
+			Headers:      headers,
+			ServiceName:  config.OTLPServiceName,
+			InstanceName: config.OTLPInstanceName,
+			Metro:        config.OTLPMetro,
+		}, slogger)
+		if err := otlpWriter.Start(ctx); err != nil {
```
Sayan-, line 153: I think this makes sense wrt parity to s2. I'd guess ~some events will race with any intermediate relay having the information required to make routing decisions. may need to reshape depending on what we want to guarantee

Sayan- (commented):
overall lgtm! left a few comments inline. broader review attached below.

part of me wonders if we should expose an enable/disable for exporting in the api but I don't think we should solve that in this PR

## Opus Semantic Review

Contract assumption: forwarding destination coupled with vm lifecycle; exporting best effort.

### A. Lifecycle & availability

| # | Property | Why | Owner | Status | Evidence |
|---|----------|-----|-------|--------|----------|
| A1 | Init failure must not crash the browser API | Optional sink shouldn't take down the VM | `main.go` wiring | **FAIL** | `main.go:145-148` `os.Exit(1)` on `Start` err. Low trigger (lazy `otlploghttp.New`), but real |
| A2 | Transient runtime export errors don't stop the sink | Endpoint flaps must self-heal | `otlpStorage`+`StorageWriter` | **PASS** | `Emit` async non-blocking; `processResult` swallows+logs (`eventsstorage.go:81`); Run returns only on ctx err |
| A3 | Read loop lives whole VM lifecycle | [1] always-on requirement | `StorageWriter.Run` | **PASS** | Exits only on `reader.Read` ctx cancel (`ringbuffer.go:146`) |
| A4 | ~Zero cost when idle | Runs on every vanilla VM | `ringBuffer.Read` | **PASS** | Parks on `<-wake` (`ringbuffer.go:145-149`) |

### B. Delivery semantics

| # | Property | Why | Owner | Status | Evidence |
|---|----------|-----|-------|--------|----------|
| B1 | Loss under pressure is safe (drops, never blocks producers or grows unbounded) | A stalled endpoint must not stall capture or OOM the VM | ring + batch queue | **PASS** | Best-effort by design (matches PR's deferred durability). `Enqueue` non-blocking, drops oldest (`batch.go:305-317`); ring evicts oldest. Just declare best-effort as the contract |
| B2 | Flush buffered records on graceful shutdown | Don't lose tail on SIGTERM | `OTLPStorageWriter.Stop` | **PASS** | `Stop`→wait `done`→`Drain`→`provider.Shutdown` (`main.go:333-339`). SIGTERM path only |

### C. Data correctness

| # | Property | Why | Owner | Status | Evidence |
|---|----------|-----|-------|--------|----------|
| C1 | Timestamps wall-clock | Suspend-skew correctness | converter | **PASS** | `time.UnixMicro(ev.Ts)`; `Ts` wall-clock contract (`event.go:79`), defaulted in `publishLocked` |
| C2 | Per-record payload bounded | Keep under relay body cap | `EventStream.Publish` | **PASS** | `truncateIfNeeded` ~1MB before ring (`eventstream.go:37`) |

### D. Auth & security

| # | Property | Why | Owner | Status | Evidence |
|---|----------|-----|-------|--------|----------|
| D1 | Authenticated export | Relay must trust the VM | `main.go` headers | **PASS** | Bearer `KERNEL_INSTANCE_JWT` (`main.go:131-133`) |
| D2 | Secrets not logged | Token leak | `config.LogValue` | **PASS** | JWT + S2 token redacted (`config.go:66-97`) |

### E. Observability & isolation

| # | Property | Why | Owner | Status | Evidence |
|---|----------|-----|-------|--------|----------|
| E1 | Export failures observable | Silent failure looks like healthy idle | `loggingExporter` + otel global logger | **FAIL** | Export errors + ring drops → app slog (no metric). **Queue-overflow drops → otel global logger, never wired (`otel.SetLogger` absent) → silently discarded** |
| E2 | OTLP failure isolated from S2 | One sink must not degrade the other | separate writers + non-blocking emit | **PASS** | Independent readers/pipelines; `Enqueue` non-blocking → stalled endpoint can't backpressure ring or S2 |

### Scorecard
- **PASS (10):** A2, A3, A4, B1, B2, C1, C2, D1, D2, E2
- **FAIL (2):** A1, E1

### Blocking Summary

A1 (don't `os.Exit`) · E1 (surface failures in logs as aggregate or by window).

rgarcia (approved):
Re-reviewed the follow-up commits: both naming findings addressed (`BTEL_OTLP_*` prefix, `InstanceJWT`/`InstanceName`/`MetroName`), sink init is now non-fatal, and export requests are byte-bounded with tests. Also verified the queue-drop logging wiring against the pinned SDK — `global.Warn` maps to logr V(1) → slog level −1, so the `LevelInfo - 1` handler admits it. LGTM.

Sayan- (approved):

server/lib/events/otlpstorage.go
```diff
@@ -0,0 +1,196 @@
...
+	sdklog "go.opentelemetry.io/otel/sdk/log"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
+)
+
+// defaultOTLPMaxBatchRecords bounds records per export request. Envelopes are
+// capped at ~1MB each at publish, so a small batch keeps typical requests
+// under a common 10MB body cap. This is a count bound, not a byte bound:
+// byte-based batching (the real guard against many large records) is deferred.
+const defaultOTLPMaxBatchRecords = 200
```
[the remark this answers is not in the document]
archandatta: Addressed in `23546c0`: the `loggingExporter` now splits each export into sub-requests under a 4MiB byte budget (`chunkBySize` / `maxOTLPExportBytes`), so request body size is bounded regardless of the 200-record count. A single record is <=1MB (the publish-time envelope cap), so it always fits one sub-request. The count cap remains only as a memory bound.

server/cmd/api/main.go
```diff
@@ -131,6 +133,36 @@ func main() {
...
+			Insecure:     config.OTLPInsecure,
+			Headers:      headers,
+			ServiceName:  config.OTLPServiceName,
+			InstanceName: config.InstanceName,
+			Metro:        config.MetroName,
+		}, slogger)
+		if err := otlpWriter.Start(ctx); err != nil {
+			// Best-effort sink: a failed exporter must not take down the browser.
+			slogger.Error("OTLP export disabled: storage writer failed to start", "err", err)
+			otlpWriter = nil
```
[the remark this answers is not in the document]
archandatta: Fixed in `df3c2cb`: `otel.SetLogger` moved into the success branch, so it only swaps the process-wide OTel logger when the sink actually starts. A failed `Start` now leaves the global logger untouched.

Issue #310, opened 2026-07-10
Copy in the live view never reaches the viewer's clipboard (implicit hosting drops clipboard sync)


breteo:
## Summary

Copying text inside a remote browser session never lands on the viewer's local clipboard, even when the live view is embedded with `allow="clipboard-read; clipboard-write"` per the docs.

## Root cause

The neko server only delivers remote-clipboard updates to the session registered as **host** (`sessions.GetHost()` — the update is silently dropped when there is no host). The chromium-headful client runs with implicit hosting, and its `remote.ts` `request()` action bails when `getters.hosting` is true — which it always is under `implicitHosting` — so **no session ever sends `control/request`, no host is ever registered, and clipboard updates are dropped for everyone**.

This is the same bug upstream neko fixed in m1k1o/neko#540 (see also m1k1o/neko#499); the client vendored here predates that fix.

## Verification (live against a production Kernel browser session)

From inside the live-view iframe of a live session:

1. Baseline: select text remotely, send a real Ctrl+C keysym chord (`keydown 65507`, `keydown 99`, …) over the data channel → **no** `control/clipboard` message ever arrives, `navigator.clipboard.writeText` never fires.
2. Send `$client.sendMessage('control/request')` once → server replies `control/locked` (session becomes host).
3. Repeat the same Ctrl+C chord → `control/clipboard` arrives with the selected text, `writeText` succeeds, and the local clipboard contains the remote selection. Copy works end-to-end.

So the server side already behaves correctly once a host exists; only the client-side request is missing.

## Fix

Port m1k1o/neko#540 to the vendored client: request control on first interaction with the screen and gate `request()` on an explicit-host `controlling` getter instead of `hosting`. PR incoming.

## Related (separate issue, mentioning for context)

macOS users pressing Cmd+C forward Meta+C to the remote Linux Chromium, which doesn't copy. The client special-cases paste (Cmd/Ctrl+V) but has no equivalent remap for copy — worth a follow-up.

PR #311, opened 2026-07-10
fix(chromium-headful): sync clipboard to viewers under implicit hosting

breteo:
## Description

Copying inside a remote browser session never reaches the viewer's local clipboard. The neko server only sends remote-clipboard updates to the registered **host** session, and under implicit hosting the vendored client never sends `control/request` (`remote.ts` `request()` bails because `hosting` is always true), so no host is ever registered and the updates are silently dropped. Full analysis and a live repro in #310.

This is a port of upstream m1k1o/neko#540 (which fixed the same bug, m1k1o/neko#499) onto the client vendored here:

- `remote.ts`: add an explicit-host `controlling` getter; gate `request()` on it instead of `hosting`.
- `video.vue`: on mouse interaction while not controlling, buffer the triggering event, send the control request, and replay the event once `controlling` flips true (`@Watch('controlling')`). Non-implicit behavior (`control-attempt` emit) is preserved.

## Verification

- `npm run build` in `images/chromium-headful/client` passes.
- Protocol change validated against a live production Kernel session: manually issuing the exact `control/request` this patch adds made the server reply `control/locked`, after which a Ctrl+C in the remote browser produced the `control/clipboard` push, `navigator.clipboard.writeText` succeeded, and the local clipboard held the remote selection — end-to-end copy working (details in #310).

---

> [!NOTE]
> **Medium Risk**
> Changes remote session control and input routing in the headful client; behavior is localized but affects who receives server clipboard and when mouse events reach the remote.
> 
> **Overview**
> Under **implicit hosting**, the client treated every viewer as “hosting” without ever sending **`control/request`**, so the neko server never registered a host and **remote clipboard updates were dropped**.
> 
> This adds a **`controlling`** getter (registered host only) and gates **`remote.request()`** on it instead of **`hosting`**. In **`video.vue`**, the first click while not controlling sends a control request, buffers the mouse down/up, and **replays** them when **`controlling`** becomes true; non-implicit mode still emits **`control-attempt`** on mousedown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 038849301103f3e6b9cedba693118dc48c58fc73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/images/chromium-headful/client/src/components/video.vue
+++ b/images/chromium-headful/client/src/components/video.vue
@@ -280,6 +280,10 @@
       return this.$accessor.connecting
     }
 
+    get controlling() {
+      return this.$accessor.remote.controlling
+    }
+
     get hosting() {
       return this.$accessor.remote.hosting
     }
@@ -804,14 +808,18 @@
       first.target.dispatchEvent(simulatedEvent)
     }
 
+    isMouseDown = false
+
     onMouseDown(e: MouseEvent) {
       this.unmuteOnInteraction()
+      this.isMouseDown = true
 
-      if (!this.hosting) {
-        this.$emit('control-attempt', e)
+      if (this.locked) {
+        return
       }
 
-      if (!this.hosting || this.locked) {
+      if (!this.controlling) {
+        this.implicitHostingRequest(e)
         return
       }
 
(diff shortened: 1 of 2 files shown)


rgarcia:
@breteo thanks so much for this and the bug report! I just merged https://github.com/kernel/kernel-images/pull/318 which should address this. This will likely go out to actual Kernel production on Monday

rgarcia:
sorry for the delayed update here but this fix is live in kernel.sh as of a few days ago (deployed July 21st ~6pm ET). Thanks again for bringing it to our attention @breteo!

breteo:
the fix works on my end! appreciate it @rgarcia!

PR #316, opened 2026-07-14
Authenticate OTLP exports with the instance JWT, refreshed on fork

archandatta:
## Summary

Keeps the in-VM OTLP exporter on the **instance JWT** (`Authorization: Bearer <KERNEL_INSTANCE_JWT>`) and resolves it **per request** so forked VMs export with a fresh credential. The relay (kernel/kernel#2763) is being changed to verify the JWT signature/expiry and read `instance_name` (+ `metro`) from its claims, then keep the existing active-session + otlp-binding lookups for authorization — claims say who's calling, the DB says what they're allowed to do.

Companion to the metro-api relay PR (kernel/kernel#2763). Wire contract: `Authorization: Bearer <instance_jwt>` + POST to `BTEL_OTLP_PATH` (set to `/otlp-relay/v1/logs` by provisioning when pointing a VM at the relay).

## Why

The instance JWT already carries `instance_name` and `metro` in its claims, so the relay gets everything a bare instance name would give it today, plus room for `session_id`/`org_id`/etc. later. The relay isn't deployed yet, so this is the cheapest moment to set the contract we actually want rather than standardizing further on bare instance names. Long term, capmonster/hcaptcha/extensionfeatures can migrate onto the same verified-JWT model.

## Fork-identity credential refresh

The exporter previously captured the credential once at process start. In fork-identity-wait mode `kernel-images-api` starts **before** identity applies and is never restarted, and the platform delivers a fresh `KERNEL_INSTANCE_JWT` in the applied identity payload — so a boot-time snapshot would export with a stale/empty credential forever. This PR resolves the identity dynamically instead:

- `otlpIdentityProvider` reads the identity (`kernel_instance_jwt` + `instance_name` + `metro`) from the applied fork-identity payload, gated on the applied marker to avoid a torn mid-apply read, falling back to boot env. It caches only a payload-sourced (or non-fork, stable) result — a pre-apply boot fallback is returned **uncached**, so the very next export after identity applies picks up the fresh values rather than waiting out a cache TTL. Outside fork mode it returns the boot identity unchanged.
- The exporter installs a `RoundTripper` (via `otlploghttp.WithHTTPClient`) that sets the bearer header from `AuthTokenFunc` on **each request**; an empty token sends no header.
- The OTLP Resource attributes (`kernel.instance_name`/`metro`) resolve via `InstanceNameFunc`/`MetroFunc` at exporter-build time — which, since export is enabled per session after identity applies, stamps the fork's own identity rather than the parent/empty boot value.

## Changes
- `server/cmd/api/main.go`: build the exporter with `AuthTokenFunc`/`InstanceNameFunc`/`MetroFunc` backed by `otlpIdentityProvider` instead of a static header map.
- `server/cmd/api/otlp_credential.go`: the dynamic identity provider (fork-payload read + boot fallback + cache-only-when-stable).
- `server/lib/events/otlpstorage.go`: `OTLPConfig.AuthTokenFunc`/`InstanceNameFunc`/`MetroFunc` + `bearerRoundTripper`.
- `server/cmd/config/config.go`: unchanged from `main` (the `InstanceJWT` field stays).

## Test plan
- `go build ./...`, `go vet`, and `go test ./cmd/config ./cmd/api/... ./lib/events` green (incl. `-race`).
- Unit: `TestOTLPIdentityProvider` (non-fork boot / fork-before-apply fallback / fork-after-apply fresh / boot fallback not cached), `TestOTLPStorageWriter_RefreshesAuthToken` (token change mid-run reflected on later exports), and `TestOTLPStorageWriter_ExportsEvents` asserting the bearer header.
- Container e2e (`server/e2e/e2e_otlp_storage_test.go`): `TestOTLPExport` asserts the bearer JWT + resource attrs land at a mock collector through a real container; `TestOTLPExportForkIdentityRefresh` boots with an empty JWT, applies a fresh one via the real `/internal/fork-identity` endpoint, and asserts exports pick up both the fresh JWT and the fork's `instance_name` with no restart.
- Full relay round-trip validated separately once kernel/kernel#2763 is deployed (its review has a live matrix against the real `XDS_JWT_SECRET`: valid→200, wrong-secret/expired/missing→401, no-session→403).

## Notes
- **Deploy ordering**: this and the relay PR set the wire contract together but neither blocks the other's review. Before a forked session exports end-to-end, the relay must verify the JWT (metro-api gets the JWT secret) and this exporter must send + refresh it. The relay isn't live yet, so there's no rollout ordering hazard.
- **Resource identity on forks**: the OTLP Resource attributes (`kernel.instance_name`/`metro`) are also resolved from the applied fork-identity payload (via `InstanceNameFunc`/`MetroFunc` read at exporter-build time), so a forked VM's telemetry is labeled with its own instance, not the parent/empty boot value. Covered by the fork-identity e2e.
- **Follow-up (not in this PR)**: under always-inject provisioning, an un-bound instance gets `401/403/404` from the relay as normal steady state (and until the relay pair lands, every export 401s). A later change can quiet those specific statuses in `loggingExporter` so they don't inflate the failure counter/logs, once the status is cleanly extractable from the `otlploghttp` error.
- **Future direction (not in this PR)**: re-mint the JWT at claim time with `session_id`/`org_id` claims via a generalized identity-update channel, and migrate the sibling relays onto the same dynamic-credential model.

---

> [!NOTE]
> **Medium Risk**
> Changes telemetry authentication and fork VM identity labeling; mis-cached or stale JWT could cause relay 401s until identity applies, but behavior is heavily tested.
> 
> **Overview**
> OTLP exports now use **`Authorization: Bearer <instance JWT>`** resolved on **each HTTP request**, matching the relay contract instead of a static header captured at startup.
> 
> **`otlpIdentityProvider`** supplies JWT, instance name, and metro from boot env in normal mode; in fork-identity-wait mode it reads the applied fork-identity payload (after the applied marker), falls back to boot until apply, and only caches stable identities so post-apply credentials show up immediately. **`OTLPConfig`** gains `AuthTokenFunc`, `InstanceNameFunc`, and `MetroFunc`; a **`bearerRoundTripper`** sets the bearer header per export.
> 
> Unit tests cover identity resolution and mid-run token refresh; container e2e verifies bearer JWT + resource attrs and fork-identity refresh without restart.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77b3f296c6d813c68e11e40832d698c5ec57cfda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (changes requested):
I'd like to go the other way here: keep the exporter on the instance JWT and correct the relay to verify it — the relay isn't deployed yet, so this pair of PRs is the cheapest moment to set the contract we actually want, rather than standardizing further on bare instance names. the JWT already carries `instance_name` (+ `metro`) in its claims, so the relay gets everything x-api-key gives it today, plus a place to put more later.

what's blocking JWT auth today — small list:

1. **relay side (kernel/kernel#2763): verify the token, keep the lookups for authorization.** plumb the JWT secret into metro-api, validate signature/expiry, take `instance_name` from the claims, and then the existing active-session + otlp-binding lookups stay exactly as they are — claims tell the relay who's calling, the DB tells it what they're allowed to do.
2. **this PR: keep sending the JWT** (`Authorization: Bearer <KERNEL_INSTANCE_JWT>`), i.e. roughly revert to the pre-PR header and keep the `InstanceJWT` config field.
3. **fork-identity-wait must refresh the exporter's credential.** the fork identity payload already delivers a new `KERNEL_INSTANCE_JWT`, but `kernel-images-api` starts before identity applies and is never restarted, and the exporter captures config once at process start — so forked VMs export with a stale/empty credential forever (true for x-api-key too, so this needs fixing regardless of which credential wins). the exporter needs to read the JWT dynamically (from the applied identity payload) rather than from env-at-boot.

future direction, briefly: re-mint the JWT at claim time with session_id / org_id / other useful instance claims, delivered through a generalized identity-update channel (the fork-identity endpoint grown into the every-VM claim-time push), and migrate capmonster/hcaptcha/extensionfeatures onto the same model. none of that blocks v1 — the boot-time JWT with the instance name claim is enough for the relay today.

cc @sjmiller609

rgarcia (approved):
re-reviewed at 1d5acf1 — this closes all three blockers from my review, and the fork handling is done better than I asked for:

- **exporter stays on the JWT** (`Authorization: Bearer`), and the credential is resolved **per request** via `bearerRoundTripper` + `AuthTokenFunc` rather than frozen into the exporter's header map at construction. the RoundTripper hygiene is right (request cloned before mutation, empty token sends no header). `KERNEL_INSTANCE_JWT` is back in config with its `LogValue` redaction restored.
- **fork-identity refresh** — `otlpIdentityProvider` reads the applied fork-identity payload (gated on the applied marker to avoid torn reads mid-apply), falls back to boot env, and never caches the pre-apply fallback — so the very next export after identity applies carries the fresh JWT, no restart. resource attrs (`kernel.instance_name` / `kernel.metro`) resolve at exporter build time, which is per export-enable — per session, after identity applies — so forked VMs stamp the fresh identity there too. I verified the build-at-Start claim against `OTLPStorageWriter.Start` and the controller's rebuild-on-enable behavior.
- **tests earn the approval on their own**: the provider unit tests pin the subtle cache rule (boot fallback must not be cached; apply picked up on the very next call despite the TTL), `RefreshesAuthToken` proves a token swap lands mid-stream, and the e2e asserts the bearer JWT + resource attrs at a real mock collector through a real container, including the fork-identity-refresh path. ran the unit suites locally — green.

one note, not blocking: deploy pairing with kernel/kernel#2763 as already discussed there — kernel/infra `XDS_JWT_SECRET` provisioning first, then the relay, then this. unset secret on the relay 401s these exports harmlessly, and the exporter's failure counter will tick until the pair lands; the follow-up to quiet expected 401/403/404s in `loggingExporter` (already noted in the PR body) is worth keeping on the list.

approving.

Sayan- (approved):
approving — the fork-identity credential refresh is done carefully (uncached pre-apply fallback, torn-read guard via the applied marker) and the e2e coverage asserts the real behavior.

one ask before merge: pls update the PR title + description. they still describe the old `x-api-key: <instance_name>` contract, but the code sends `Authorization: Bearer <instance_jwt>` (and `instanceJWTProvider` is now `otlpIdentityProvider`). the stale metadata is what keeps regenerating the "auth contract mismatch" bot threads on both this PR and kernel/kernel#2763.

PR #330, opened 2026-08-10
Bypass private networks by default

hiroTamada:
## Summary

- launch Chromium with a default proxy bypass list for RFC1918, CGNAT, and IPv6 ULA ranges when no bypass list is configured
- preserve existing base and runtime flag ordering so custom and explicitly empty bypass lists remain authoritative
- cover default, replacement, clearing, and base/runtime precedence

## Why

Browsers using an egress proxy should retain direct access to standard private network destinations without requiring a runtime Chromium restart. Runtime configuration remains authoritative when callers need a custom list or no bypasses.

## Testing

- `go test -race $(go list ./... | grep -v /e2e$)`
- `go vet ./...`

---

> [!NOTE]
> **Low Risk**
> Launcher-only Chromium flag wiring with explicit opt-out via existing bypass-list flags; no auth or data-path changes.
> 
> **Overview**
> Chromium in the kernel image now gets a **default `--proxy-bypass-list`** for RFC1918, CGNAT (`100.64.0.0/10`), and IPv6 ULA (`fc00::/7`) so traffic to private networks can go direct when an egress proxy is in use, without a runtime restart.
> 
> `chromium-launcher` applies this via `withDefaultPrivateNetworkBypass` **after** `CHROMIUM_FLAGS` and runtime overlay are merged. If the merged flags already include `--proxy-bypass-list` (including `=` empty or bare flag), the image default is **not** added—callers keep full control. Unit tests cover default injection, custom/empty override, and that an existing configured bypass is not duplicated when the default token also appears in the merged slice.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e874dd8d7f35ad6e60e70bdea61c5d4808199c99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/cmd/chromium-launcher/main.go
+++ b/server/cmd/chromium-launcher/main.go
@@ -80,7 +80,8 @@ func main() {
 		fmt.Fprintf(os.Stderr, "failed reading runtime flags: %v\n", err)
 		os.Exit(1)
 	}
-	final := mergeChromiumFlags(baseFlags, runtimeTokens)
+	final := chromiumflags.MergeFlagsWithRuntimeTokens(baseFlags, runtimeTokens)
+	final = withDefaultPrivateNetworkBypass(final)
 
 	// Diagnostics for parity with previous scripts
 	fmt.Printf("BASE_FLAGS: %s\n", baseFlags)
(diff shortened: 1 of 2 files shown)


rgarcia (approved):
reviewed — baking the standard private ranges into the image is the right approach. it preserves pre-proxy-v3 direct routing for existing integrations without adding a claim-time Chromium restart, while custom and explicitly empty runtime lists remain overrides.

### override precedence

- `server/cmd/chromium-launcher/main.go:163-173` — filtering every configured token equal to `defaultPrivateNetworkBypassFlag` can move an explicit runtime override ahead of an older base override, changing which list wins.

  For example:

  ```go
  baseFlags := "--proxy-bypass-list=preview.internal"
  runtimeTokens := []string{defaultPrivateNetworkBypassFlag}
  ```

  `MergeFlagsWithRuntimeTokens` correctly produces:

  ```text
  preview.internal, standard-default
  ```

  so the runtime standard list is last and authoritative. The new filtering removes that final token and returns:

  ```text
  standard-default, preview.internal
  ```

  which makes the stale base list win.

  Avoid relocating configured occurrences. The simplest fix is to prepend the image default and retain all configured tokens; an identical duplicate is harmless and preserves Chromium’s last-token precedence. Alternatively, inject the image default only when `configured` does not already contain it, while leaving the configured sequence untouched.

  Add a regression case in `server/cmd/chromium-launcher/main_test.go:18-53` where a runtime token equal to `defaultPrivateNetworkBypassFlag` follows a different base bypass list, and assert that the standard list remains last.

The unit tests otherwise cover the expected default, custom replacement, and explicit clearing order. Actual direct routing and empty-list clearing should still be verified against the built image by the corresponding real-browser e2e in kernel/kernel#3233.

PR #55, opened 2025-08-16
[Extending API] /clipboard

raiden-staging:
- Extending server with `/clipboard` routes

```bash
# Clipboard
# | GET /clipboard - Get clipboard content
curl http://localhost:10001/clipboard
# | > {"type":"text","text":"lorem_ipsum"}

# | POST /clipboard - Set clipboard content
curl -X POST -H "Content-Type: application/json" \
  --data '{"type":"text","text":"hello"}' \
  http://localhost:10001/clipboard

# | GET /clipboard/stream - Stream clipboard changes as SSE
# Keep this running in one terminal
curl -N http://localhost:10001/clipboard/stream

# In another terminal, set clipboard content to see events
curl -X POST -H "Content-Type: application/json" \
  --data '{"type":"text","text":"this_notifies_stream"}' \
  http://localhost:10001/clipboard
```

---

## TL;DR

Introduced new `/clipboard` API endpoints to get, set, and stream system clipboard content, extending server capabilities for external application interaction.

## Why we made these changes

To enhance server functionality by providing programmatic access to the system clipboard, enabling external applications to interact with clipboard data.

## What changed?

*   **Added Clipboard API:** Implemented GET, POST, and SSE stream endpoints for clipboard operations in `clipboard.go`, leveraging `xclip`.
*   **API Specification & Generation:** Updated `openapi.yaml` and regenerated OpenAPI files (`oapi.go`, `clipboard_types.go`, `server_handlers.go`).
*   **Server Integration:** `ApiService` was enhanced with a `clipboardManager`, and `main.go` explicitly registers the new handlers.
*   **Dependencies:** `go.mod` and `go.sum` were updated to reflect new direct and indirect Go dependencies, including `oapi-codegen`.
*   **Testing:** New unit tests were added in `clipboard_test.go` for the clipboard API functionalities.
*   **Documentation:** `README.md` was updated with `curl` examples for the new routes.

<sup>_Description generated by Mesa. [Update settings](https://app.mesa.dev/onkernel/settings/pull-requests)_</sup>

PR #332, opened 2026-08-12
Add captcha task-start and challenge-result events

archandatta:
## Summary

- add task-level `captcha_solve_started` and challenge-level `captcha_challenge_result` telemetry
- distinguish challenge outcomes as `solved`, `failure`, `timeout`, or `abandoned`
- require `data` on both new events and `duration_ms` on challenge results
- preserve the existing generated captcha type and solve-result status symbols for downstream Go consumers

## Contract

`captcha_solve_started` records that a solver accepted one task. `task_id` pairs that start with its terminal `captcha_solve_result`, while `challenge_id` groups every task from one visible challenge. Delivery remains best-effort and unordered, so event arrival or absence does not indicate current solve state.

`captcha_challenge_result` is emitted once for the visible challenge. `failure` means the producer reached an explicit terminal solver failure or exhausted its attempts while the challenge remained. `timeout` is reserved for an actual challenge-level wait budget. `abandoned` covers observation ending without either a solved signal or an explicit terminal solver outcome.

The shared captcha taxonomy keeps the legacy generated `BrowserCaptchaSolveResultEventDataCaptchaType` name and constants through a schema alias. Existing relay code therefore compiles unchanged when it updates this module.

## Rollout

Schema and generated image API types only; no producer behavior changes here. The public API schema and extension producer are updated in their corresponding PRs before consumers rely on the new challenge event.

## Testing

- `go test ./lib/oapi ./lib/events/...` — pass, including generated-symbol and required-field compatibility coverage
- `go build ./...` — pass
- `go vet ./...` — pass
- OpenAPI generation run twice — byte-identical generated output
- `go test ./lib/capmonsterrelay` on `kernel/kernel#3432` with this module replacement — pass without relay changes
- GitHub `test-server-e2e` — all cases except `TestReplayRecordingIncludesAudioTrack`; that unrelated test failed twice because the generated audio track ended 0.38s / 1.29s before its recording-duration threshold. The base branch has a green full e2e run on the same base commit.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 1183b8d6f7bd248a6237e89ff3e3ef480369968d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sayan- (commented):
Wire format is additive and safe: the new types land in a taxonomy-only union and `challenge_id` is optional. Two things I'd still fix.

1. Build break downstream. This deletes `BrowserCaptchaSolveResultEventDataCaptchaType` and its constants, and renames the status constants from `Success`/`Failure`/`Timeout`/`Abandoned` to prefixed forms. `kernel/kernel` uses both in `packages/metro-api/lib/capmonsterrelay/relay.go`, line 332 for status and line 363 for `normalizeCaptchaType`'s signature and returns. metro-api resolves that import through a `replace` onto the internal mirror, so it breaks on the next sync plus module bump rather than at merge. Worth a companion PR landed first.

The status rename also looks unintentional. oapi-codegen only dropped the prefix because nothing collided; removing the captcha-type enum changed the collision set.

2. `captcha_challenge_result.status` is `solved`/`abandoned` only, so a solver hard-failure and a dismissed widget are indistinguishable. Adding `failure` or `timeout` later means widening a shipped SDK enum, which is the thing the rest of this PR is careful to avoid.

Nits: `data` is optional on the event, so required `challenge_id` isn't enforced. `duration_ms` is optional here but required on the task result.

masnwilliams (approved):
this now models the lifecycle correctly:

- `captcha_solve_result` remains task-level and may occur once per image-grid round.
- `task_id` correlates each task start/result pair.
- `challenge_id` groups those task attempts.
- `captcha_challenge_result` provides the once-per-visible-challenge terminal signal, with `solved`, `failure`, `timeout`, and `abandoned`.

the latest revision also preserves the existing generated Go API, requires data on the new events, requires challenge duration, and adds compatibility coverage. no remaining structural code-quality blockers from me.

this PR is schema-only, so the behavior is not complete until the producer and consumer changes land: kernel-images-private #339, kernel #3418, and managed-auth-cua-concept #55.

the remaining red e2e check is the unrelated recording-audio-duration assertion; the relevant unit, build, vet, generation, BugBot, and image-build checks pass.

Sayan- (approved):
Confirming all feedback addressed. Thanks for iterating!

PR #348, opened 2026-08-22
Enable WebMCP by default and merge feature-list flags

rgarcia:
## Summary

Enables WebMCP (Chromium's `navigator.modelContext` API plus its CDP domain) in default browser launch flags, and makes flag merging feature-list-aware so per-session flags can't silently drop image defaults.

### Feature strings

Verified against the Chromium 151 sources: `chrome://flags/#enable-webmcp-testing` maps to base feature **`WebMCPTesting`** (disabled by default; implies the runtime `WebMCP` feature in the renderer), and `devtools-webmcp-support` maps to **`DevToolsWebMCPSupport`**, which gates registration of the CDP `WebMCP` domain handler. Browser-side use also requires the `Tools` permissions policy (`EnableForSelf`, depends on WebMCP), which top-level pages get by default.

On Chromium 151+, `--enable-features=WebMCPTesting,DevToolsWebMCPSupport` turns on both the JS API and the WebMCP CDP domain.

### What changed

- Headless default flags (`server/cmd/wrapper/chromium.go`) and headful defaults (`images/chromium-headful/run-{unikernel,docker}.sh`): add `--enable-features=WebMCPTesting,DevToolsWebMCPSupport`.
- `chromiumflags.MergeFlags`: union `--enable-features` / `--disable-features` CSV values across base and runtime sources into single tokens. Chromium keeps only one value per switch, so previously a session-supplied `--enable-features=X` would have silently dropped the image's entire default list (or vice versa). A feature present in both lists resolves to disabled inside Chromium, so sessions can still opt out with `--disable-features=WebMCPTesting`.
- `chromiumflags.TranslateKernelDisableFeatures`: folds a kernel-namespaced `--kernel-disable-features=` token into the single `--disable-features` switch at flag-assembly time (both in the launcher and before `/chromium/flags` is written). Chromium ignores unknown switches, so on images predating translation the token is inert — an opt-out degrades to a no-op instead of last-win clobbering boot-time disable lists.

### Tradeoffs

- Fingerprint: enabling by default exposes `navigator.modelContext` to pages, which stock Chrome 151 does not show without an origin trial or experimental-web-platform-features. Sites can detect that surface as a deviation. Callers that need stock-matching behavior can pass `--disable-features=WebMCPTesting,DevToolsWebMCPSupport` (or the kernel-namespaced token) through existing flag mechanisms; the control plane does this for stealth sessions.

### Testing

- `go test ./lib/chromiumflags/` — extended table tests for feature-list merging (union across sources, opt-out via disable list, coexistence) and for kernel-token translation (fold into existing list, create list when absent, idempotency). All pass.
- `go build ./...` and full `go test ./...` run; remaining failures (display/OTLP/extension e2e tests requiring a running Chromium/X environment) fail identically on main.

### DIFF
--- a/server/lib/chromiumflags/chromiumflags.go
+++ b/server/lib/chromiumflags/chromiumflags.go
@@ -86,10 +86,7 @@ func union(base, rt []string) []string {
 // (base/feature_list.cc), which terminates the name at the first of ':',
 // '.', or '<' — parameter, group, and study separators respectively.
 func canonicalEnableFeatureName(entry string) string {
-	if i := strings.IndexAny(entry, ":.<"); i >= 0 {
-		return entry[:i]
-	}
-	return entry
+	return stripLeadingStar(cutAtFirst(entry, ":.<"))
 }
 
 // canonicalDisableFeatureName strips decoration from a --disable-features
@@ -97,10 +94,22 @@ func canonicalEnableFeatureName(entry string) string {
 // RegisterOverridesFromCommandLine, which only splits on '<'; dots and colons
 // are part of the name.
 func canonicalDisableFeatureName(entry string) string {
-	if i := strings.IndexByte(entry, '<'); i >= 0 {
+	return stripLeadingStar(cutAtFirst(entry, "<"))
+}
+
+// cutAtFirst returns entry up to its first byte in separators, if any.
+func cutAtFirst(entry, separators string) string {
+	if i := strings.IndexAny(entry, separators); i >= 0 {
 		return entry[:i]
 	}
 	return entry
+}
+
+// stripLeadingStar removes the optional '*' default-reset prefix that
+// RegisterOverride consumes before inserting into its override map, so
+// *Foo and Foo collapse to one canonical key.
+func stripLeadingStar(name string) string {
+	return strings.TrimPrefix(name, "*")
 }
 
 // mergeFeatureEntries merges base and runtime feature-list entries into one
--- a/server/lib/chromiumflags/chromiumflags.go
+++ b/server/lib/chromiumflags/chromiumflags.go
@@ -36,10 +36,11 @@ func appendCSVInto(dst *[]string, csv string) {
 	}
 }
 
-// parseTokenStream extracts extension-related flags and collects non-extension flags.
-// It returns the list of non-extension tokens and, via references, fills the buckets for
-// --load-extension, --disable-extensions-except and a possible --disable-extensions token for that stream.
-func parseTokenStream(tokens []string, load, except *[]string, disableAll *string) (nonExt []string) {
+// parseTokenStream extracts extension-related and feature-list flags and collects the rest.
+// It returns the leftover tokens and, via references, fills the buckets for
+// --load-extension, --disable-extensions-except, --enable-features, --disable-features
+// and a possible --disable-extensions token for that stream.
+func parseTokenStream(tokens []string, load, except, enableFeatures, disableFeatures *[]string, disableAll *string) (nonExt []string) {
 	for _, tok := range tokens {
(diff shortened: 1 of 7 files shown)


server/lib/chromiumflags/chromiumflags.go
```diff
@@ -128,29 +135,52 @@ func ReadOptionalFlagFile(path string) ([]string, error) {
...
 
 	// Merge extension lists - include paths from --disable-extensions-except in load paths
 	// since we no longer emit --disable-extensions-except
 	mergedLoad := union(baseLoad, rtLoad)
 	mergedLoad = union(mergedLoad, baseExcept)
 	mergedLoad = union(mergedLoad, rtExcept)
 
+	// Merge feature lists - a feature listed in both enable and disable stays in both;
+	// Chromium resolves that conflict as disabled.
+	mergedEnableFeat := union(baseEnableFeat, rtEnableFeat)
```
masnwilliams: `union` deduplicates the full entry string, but Chromium feature entries may include trial/parameter decorations. For example, a base `Feature<TrialA` and runtime `Feature<TrialB` survive as two entries even though Chromium resolves both as `Feature`; Chromium then keeps the first configuration, so the runtime override can silently lose. Please merge entries by canonical feature name and let the runtime entry replace the base entry, with a regression test for decorated variants. Since `MergeFlags` now handles arbitrary feature-list flags, exact-string deduplication is not sufficient.

masnwilliams (changes requested):
The overall approach looks right: enabling WebMCP in the image, merging feature-list switches to avoid Chromium's last-switch-wins behavior, and using a namespaced disable switch for mixed-version rollout all make sense.

There is one correctness issue in the generic feature merger that should be fixed before shipping; I left it inline. Please also add a regression test covering decorated entries for the same feature name.

Non-blocking follow-ups: returning a structured parse result would be easier to maintain than six output parameters, and image-level coverage should verify that WebMCP is available by default and absent after the stealth opt-out.

server/lib/chromiumflags/chromiumflags.go
```diff
@@ -74,6 +81,60 @@ func union(base, rt []string) []string {
 	return out
 }
 
+// canonicalEnableFeatureName strips trial/parameter decoration from an
+// --enable-features entry. Those go through ParseEnableFeatureString
+// (base/feature_list.cc), which terminates the name at the first of ':',
+// '.', or '<' — parameter, group, and study separators respectively.
+func canonicalEnableFeatureName(entry string) string {
```
masnwilliams, line 88: Please also strip an optional leading `*` from the canonical key in both enable and disable paths. `RegisterOverride` removes that prefix before inserting into `overrides_` (`*Foo` means `OVERRIDE_USE_DEFAULT`), so Chromium treats `*Foo` and `Foo` as the same feature. Today base `*Foo` plus runtime `Foo` emits both in that order; the base entry registers as `Foo` first and the runtime override is ignored. The inverse ordering fails for the same reason. Add regression cases for starred/unstarred replacement in both switch buckets.

masnwilliams (changes requested):
The decorated-feature handling is now correct for the enable/disable grammars, and the new tests cover the separators well. One canonical-identity case remains: Chromium strips a leading `*` in `RegisterOverride` before indexing the override, so starred and unstarred entries also need to collide in this merger. I left the case inline.

masnwilliams (approved):
The original blocker is fixed correctly. The remaining leading-asterisk case is valid but non-blocking: Kernel does not currently emit starred feature entries, the WebMCP path does not use them, and this change does not regress the intended rollout. Treating that edge case as follow-up hardening rather than holding this release. Tests and CI are green.

PR #356, opened 2026-08-25
Memoize Playwright page target IDs

rgarcia:
## Summary

- cache successful Playwright page-to-CDP target ID lookups in a WeakMap
- coalesce concurrent discovery and prevent pre-reset work from repopulating the cache
- refresh cached IDs on resolver retries and reset them on browser disconnect
- build a per-attempt pageByTargetId index for the tab-to-page join
- serve each resolution attempt with a single browser-level CDP session (the active-tab listing and every tab-to-page join share one session instead of opening one per active tab)
- add focused Node tests for reuse, refresh, failures, concurrency, and reset behavior

## Stack

Based on #347. Replaces #352/#355 after the `resolve_active_page` request option was dropped from the stack.

## Validation

- `make test-runtime` (9/9)
- `go vet ./...`
- Playwright daemon esbuild bundle build

### DIFF
--- a/.github/workflows/server-test.yaml
+++ b/.github/workflows/server-test.yaml
@@ -36,6 +36,11 @@ jobs:
           cache: true
           cache-dependency-path: server/go.sum
 
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
       # categorygen's checks (unclassified route, category that isn't control or
       # platform, classified route with no handler) only run when the generator
       # does, and its only other caller is `make oapi-generate`, which needs the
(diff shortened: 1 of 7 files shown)


masnwilliams (approved):
reviewed the stacked layer at `16d6b3a`. one root CDP session per attempt removes the nested session churn, while the weak cache is refreshed on retry and reset on browser disconnect/timeout. the cache tests cover reuse, failed discovery, refresh, in-flight work, and reset.

validation beyond BugBot: runtime tests and the full non-e2e race suite pass; both Dockerfiles pass build checks; the bundled daemon passed the deterministic cross-context case, 30 rapid cross-origin transitions with no failures/wrong pages, page recreation, and an unmatched raw-CDP tab. with 41 pages, warm active-first/active-last resolution was ~10ms in both positions.

this PR targets the lower stack branch, so the main-targeted image/server workflows have not run on this layer yet. please let those run after #347 merges and this retargets to `main`, before merging/releasing. approved subject to that normal CI gate.

PR #363, opened 2026-09-02
Avoid clipboard prompts in embedded live views

masnwilliams:
## summary

- avoid requesting clipboard-read permission from embedded live views during focus and mouse entry
- continue proactive clipboard sync when an embed already has permission
- leave top-level live-view clipboard behavior unchanged

## why

A clipboard permission prompt can take keyboard focus from a cross-origin live-view iframe while mouse input continues through its independent handler. The overlay remains the active element inside the iframe, but trusted key events no longer reach it.

The native paste path already reads `ClipboardEvent.clipboardData`, so embedded users can still paste without triggering a permission request.

## validation

Tested against exact image `kernel-cu-v63:863808b` using Chromium 151, fresh viewer profiles, trusted OS-level input, client-stage counters, privacy-safe Neko receipt/injection markers, and a remote input assertion.

- baseline: clipboard permission remained `prompt`, iframe focus dropped, client sent the mouse down, Neko received/authorized/injected both pointer events, and 0/5 keys reached the client or server
- patched: iframe focus remained active, 5/5 keys reached the client, Neko received/authorized/injected all 10 key down/up packets, and the remote input received all five characters
- direct top-level control: 5/5 fresh-profile runs retained focus and delivered all keys, confirming this change remains scoped to the embedded failure
- native paste delivered the exact fixture through the embedded live view
- `bun test tests` (4 passed)n- `bun run lint --no-fix` (0 errors; existing warnings)
- `bun run build` (passes; existing lint and bundle-size warnings)

---

> [!NOTE]
> **Low Risk**
> Single guarded change to optional clipboard polling in the Neko video overlay; no auth or server-side impact, with top-level behavior preserved.
> 
> **Overview**
> **`syncClipboard`** in the headful video client no longer calls **`navigator.clipboard.readText()`** from cross-origin embedded live views unless **`clipboard-read`** is already **granted**, which stops the browser permission prompt from stealing keyboard focus while mouse input still works.
> 
> The method now bails out early when clipboard read is unavailable or the document lacks focus, then—in **iframe** contexts only—queries **`navigator.permissions`** and returns without reading when permission is not granted. **Top-level** live views keep the previous proactive sync path; embeds that already have permission still sync on focus and mouse enter. Paste via **`ClipboardEvent.clipboardData`** is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4cdbedb07528fb23387af5b3b790ef29810d4ad4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (approved):
looks good — embedded clipboard reads require existing permission, top-level behavior stays unchanged, and native paste remains supported.

optional: add regression tests for granted, prompt/denied, and unsupported permission queries.

reviewed the code; browser validation, lint, and build results are from the PR description, not independently rerun.

PR #366, opened 2026-09-03
Add bounded scale-to-zero leases

tnsardesai:
## problem

A handler that streams a large response can finish once the final bytes are accepted by the guest TCP send buffer, before those bytes have reached metro-api. Releasing the request-scoped scale-to-zero hold at that point can suspend the VM with response bytes still queued.

Metro-api needs a crash-safe way to hold scale-to-zero disabled until the guest response has been fully read. The existing disable/enable pin is intentionally a permanent boolean used by lifecycle management, so reusing it for individual responses would let one response release another caller's pin and could leave a VM pinned forever if metro-api exits.

## change

- add additive `PUT`/`DELETE /scaletozero/leases/{lease_id}` endpoints
- keep named leases independent from the existing permanent pin and request refcount
- renew a repeated lease ID idempotently
- expire leases automatically after a caller-supplied TTL, capped at five minutes by the API, with a safe default scheduler for directly constructed controllers
- cap each guest at 1,024 concurrent leases and return a distinct 409 conflict at the limit
- log background re-enable failures and keep explicit release retries effective
- re-enable scale-to-zero only after the request refcount, permanent pin, and all bounded leases are gone
- regenerate the OpenAPI server/client and telemetry operation map

Existing `/scaletozero/disable` and `/scaletozero/enable` behavior is unchanged.

## tests

- `make oapi-generate`
- `go build ./...`
- `go vet ./...`
- `go test -race ./lib/scaletozero ./cmd/api/api`
- `go test ./lib/scaletozero -run '^TestDebouncedControllerLeaseRenewalSurvivesStaleExpiry$' -count=50`
- `go test ./lib/scaletozero -run '^TestDebouncedControllerLeaseDefaultsExpiryScheduler$' -count=50`
- `go test -race $(go list ./... | grep -v '/e2e$')` (all changed packages passed; one unrelated `lib/devtoolsproxy` Chromium temp-directory cleanup raced once)
- `go test -race ./lib/devtoolsproxy` (retry passed)

## rollout

Release this in browser images before relying on bounded leases from metro-api. The metro-api change treats a 404/405 as an older image and retains its independent stall-wake watchdog during the image rollout.

---

> [!NOTE]
> **Medium Risk**
> Changes when VMs can scale to zero (infrastructure lifecycle) but the API is additive, TTL-bounded, and existing disable/enable pin behavior is unchanged.
> 
> **Overview**
> Adds **bounded scale-to-zero leases** so callers (e.g. metro-api) can keep a VM awake for a limited time after a streaming response ends, without using the permanent pin or request refcount.
> 
> New **`PUT` / `DELETE /scaletozero/leases/{lease_id}`** endpoints validate lease IDs and `ttl_seconds` (1–300), delegate to `AcquireLease` / `ReleaseLease` on the existing scale-to-zero controller, and map lease-cap exhaustion to **409**. OpenAPI client/server and telemetry operation categories are regenerated.
> 
> **`DebouncedController`** now tracks named leases with timers (renew on repeat ID), a **1024-lease cap**, and treats active leases like pins/refcounts when deciding re-enable; cooldown re-enable failures are logged instead of silently ignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d0b726da8ffc11eb3627e7cd91c1bdb53eb9f8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/lib/scaletozero/scaletozero.go
```diff
@@ -213,10 +227,71 @@ func (c *DebouncedController) Unpin(ctx context.Context) error {
...
+
+func (c *DebouncedController) expireLease(id string, expired *lease) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.leases[id] != expired {
+		return
+	}
+	delete(c.leases, id)
+	_ = c.maybeReenableLocked(context.Background())
```
tnsardesai: G1: the `Enable` error is dropped with no log. With no cooldown, a failed write here leaves `disabled=true` with zero holders until some later request happens to re-drive `maybeReenableLocked` — a VM that never auto-standbys with nothing in the guest logs explaining why. Log at error with `lease_id`, and consider a retry timer like the cooldown path.

server/cmd/api/api/scaletozero.go
```diff
@@ -22,3 +26,26 @@ func (s *ApiService) EnableScaleToZero(ctx context.Context, _ oapi.EnableScaleTo
 	}
 	return oapi.EnableScaleToZero204Response{}, nil
 }
+
+func (s *ApiService) AcquireScaleToZeroLease(ctx context.Context, req oapi.AcquireScaleToZeroLeaseRequestObject) (oapi.AcquireScaleToZeroLeaseResponseObject, error) {
```
tnsardesai: G2: nothing bounds the number of distinct lease IDs. A metro-api bug that mints a fresh ID per renew would pin the VM for 5 minutes past the last call and grow the map without limit. Cap the live lease count (e.g. 64) and reject above it.

server/lib/scaletozero/scaletozero.go
tnsardesai: fixed: expiry failures now log at error with the lease ID, and cooldown-timer failures are logged as well. explicit release remains retryable, with a new test covering the failed-Enable then retry path.

server/cmd/api/api/scaletozero.go
tnsardesai: fixed: the controller rejects a new lease after 1,024 live leases while still allowing renewal of existing IDs. the limit path is tested to ensure it does not change the underlying scale-to-zero state.

server/lib/scaletozero/scaletozero.go
```diff
@@ -213,10 +230,78 @@ func (c *DebouncedController) Unpin(ctx context.Context) error {
...
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.leases == nil {
+		c.leases = make(map[string]*lease)
+	}
+	previous := c.leases[id]
+	if previous == nil && len(c.leases) >= maxScaleToZeroLeases {
+		return fmt.Errorf("scale-to-zero lease limit reached")
```
tnsardesai: G4 (nit): this surfaces as a 500 from the handler, indistinguishable from a control-file write failure. Harmless now that metro-api falls back to the watchdog on any pin error; a sentinel error → 429/409 is the cheap upgrade if the distinction ever matters.

server/lib/scaletozero/scaletozero.go
```diff
@@ -213,10 +235,78 @@ func (c *DebouncedController) Unpin(ctx context.Context) error {
...
+		if err := c.ctrl.Disable(ctx); err != nil {
+			return err
+		}
+		c.disabled = true
+	}
+	if previous != nil {
+		previous.timer.Stop()
+	}
+	current := &lease{}
+	current.timer = c.scheduleLeaseExpiry(ttl, func() {
```
tnsardesai: I1 (nit): only the two constructors set `scheduleLeaseExpiry`; a struct-literal or zero-value `DebouncedController` nil-derefs here. A nil fallback to `time.AfterFunc` at the top of `AcquireLease` (under `c.mu`) closes that.

sjmiller609 (approved):
- **server/lib/scaletozero/scaletozero.go:71-75** — Considering returning an error if called in Hypeman instead of succeeding.
- **server/openapi.yaml:1647-1698** — consider if we should delete the "release" endpoint, since we could just renew the lease with a shorter duration? Seems subjective though, so whatever you prefer is good (e.g. just set TTL to 1s or 0s versus DELETE)
- note: I checked on the clock jumping concern and I think there isn't one because this uses relative monotonic timing

Sayan- (changes requested):
blocking since it sounds like we prefer #369

PR #386, opened 2026-09-11
Bundle inert Docker runtime in browser images

rgarcia:
## summary

- bundle the Docker 29.1.3 CLI, daemon, and required container runtime binaries in headful and headless browser images
- keep the daemon inert by leaving it out of supervisor and omitting service units
- configure `fuse-overlayfs` as the default storage driver and select `iptables-legacy` for bridge networking
- validate the daemon configuration during image builds

## testing

- built the Chromium headful image
- built the Chromium headless image
- verified a normal browser image launch does not start `dockerd`
- started the bundled daemon in a Hypeman-backed browser and confirmed `fuse-overlayfs`, cgroup v2, nested Alpine execution, default bridge routing, and outbound HTTPS


shared/docker/daemon.json
```diff
@@ -0,0 +1,3 @@
+{
+  "storage-driver": "fuse-overlayfs"
+}
```
sjmiller609, line 3: No action requested:

Hypeman can run native overlay2 when Docker’s data root is backed directly by ext4; FUSE substantially penalizes metadata-heavy pulls and builds.

I checked on this, but this is not practical in this situation because we wouldn't run a dedicated storage mount for docker. So just be aware of IO performance issues from fuse.

shared/docker/install-docker.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
sjmiller609, line 1: nit: this file is called install docker but does not actually install docker

sjmiller609 (approved):

shared/docker/install-docker-dependencies.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
rgarcia: Renamed this to `install-docker-dependencies.sh` and updated both Dockerfiles.

PR #389, opened 2026-09-12
Fix nested Docker exec root isolation

rgarcia:
## summary

- build and ship a dedicated `runc-hypeman` runtime that restores the container init process root after entering its namespaces
- configure nested Docker to use the patched runtime with `fuse-overlayfs`
- add a public installer that rebuilds the same pinned runtime from source on older browser images
- add a focused Hypeman CI regression covering image and writable-layer visibility across `docker exec`

## why

Hypeman browser images run inside an outer chroot. Upstream `runc exec` joins the container namespaces without resetting the calling process root, so exec processes see the Hypeman initrd instead of the target container filesystem.

The patch opens the container init process root before namespace transitions, then uses that descriptor to enter the correct root after `setns`. Stock `runc` remains installed separately.

## testing

- first Hypeman CI run reproduced the bug: container PID 1 reported Alpine 3.20.10 while `docker exec` reported the Hypeman initrd's Alpine 3.22.5
- latest Hypeman regression passes with the patched runtime
- headful and headless image builds pass
- server unit and e2e suites pass
- local static runtime build and shell/config syntax checks pass

### DIFF
--- a/.github/workflows/server-test.yaml
+++ b/.github/workflows/server-test.yaml
@@ -62,6 +62,38 @@ jobs:
         run: make test-runtime
         working-directory: server
 
+  test-hypeman-nested-docker:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    needs: [build-headless]
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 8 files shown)


sjmiller609 (commented):
let's consider addressing this in a hypeman-generic way instead of one off here.

https://github.com/kernel/hypeman/pull/478

PR #392, opened 2026-09-14
Harden live-view read-only toggles and acknowledge mode changes

ehfeng:
Make live-view input locking reliable without reconnecting the viewer. The client tells its parent whether it supports mode changes and acknowledges each applied change, allowing the dashboard to keep the iframe loaded and fall back to a reload for older images.

The viewer validates the sending parent window and origin, preserves read-only mode across connection resets, and restores the server's configured control behavior when unlocked. Locking releases held input and control, and blocks clipboard and keyboard modifier writes—including clipboard reads and paste shortcuts already in progress. Existing messages without a request ID remain supported.

Validation: 22 tests pass after a fresh dependency install, and the production client build succeeds. The viewer was live-tested on an isolated browser session: playback continued across lock/unlock, input was blocked while locked, held keys and buttons were released, unrelated-window messages were rejected, and iframe navigation/WebSocket counts stayed unchanged. Modifier resynchronization is covered by regression tests.

Issue: [KERNEL-1948](https://linear.app/onkernel/issue/KERNEL-1948/toggle-session-live-view-read-only-mode-without-reloading). Companion dashboard change: [kernel#3998](https://github.com/kernel/kernel/pull/3998).

---

> [!NOTE]
> **Medium Risk**
> Changes remote input locking, clipboard forwarding, and cross-origin postMessage handling for embedded viewers; behavior is heavily tested but affects security-sensitive control paths.
> 
> **Overview**
> **Hardens embedded live-view read-only mode** so parents can toggle input lock via `postMessage` without reloading the iframe, with explicit capability advertisement and per-request acknowledgement for the dashboard.
> 
> Read-only state moves from a local override in `app.vue` into the **remote Vuex store** (`setReadOnly`), which locks input, disables implicit hosting while preserving the server’s configured hosting on unlock, and keeps read-only across transport `reset`s. Parent `KERNEL_SET_READ_ONLY` messages are validated through **`isReadOnlyMessage`** (exact parent window, non-opaque referrer origin, typed payload); valid toggles release remote control and **`releaseInput`** on the video layer, then reply with **`KERNEL_READ_ONLY_CHANGED`**. **`KERNEL_CONNECTED`** now includes `capabilities: ['setReadOnly']`.
> 
> The **video** layer treats read-only as locked input: guards clipboard sync/paste shortcuts during async work, tracks pressed mouse buttons for cleanup, and avoids caching keyboard modifier state while locked.
> 
> CI for the live-view client adds **Node 22 + `npm ci`** before Bun tests; a new **README** documents the embed contract. **Unit tests** cover message validation, store behavior, and input edge cases during lock transitions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 587dadff8a5174730f26f690c39f68bd166c2ee6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


images/chromium-headful/client/src/components/video.vue
```diff
@@ -880,7 +886,7 @@
     }
 
     onMouseEnter(e: MouseEvent) {
-      if (this.hosting) {
+      if (this.hosting && !this.locked) {
```
masnwilliams, line 889: Blocking synchronization here while locked is correct, but `onMouseLeave` still overwrites `keyboardModifierState` with the local state during that same lock. After unlock, `syncKeyboardModifierState` sees the cached values as already synchronized and sends nothing, so Caps/Num/Scroll changes made while read-only can remain unapplied remotely. I reproduced this transition locally. Please keep observed local state separate from last-sent remote state, or invalidate/force a remote sync when unlocking.

images/chromium-headful/client/tests/video-input.test.ts
```diff
@@ -0,0 +1,133 @@
+import { describe, expect, test } from 'bun:test'
+import { readFileSync } from 'node:fs'
+import { runInNewContext } from 'node:vm'
+import ts from 'typescript'
+import { parseComponent } from 'vue-template-compiler'
```
masnwilliams, line 5: These new imports make the required `test-live-view` check fail before tests run: `.github/workflows/live-view-test.yaml` executes `bun test tests` without installing the client dependencies, so `typescript`, `vue-template-compiler`, and `typed-vuex` cannot be resolved. The 17 tests pass after `npm ci`; please add the dependency-install step to the workflow so CI exercises them.

masnwilliams (changes requested):
requesting two targeted changes: preserve modifier synchronization across read-only transitions, and make the live-view test workflow install the dependencies now imported by the test suite. the parent-message hardening, acknowledgement protocol, input release, and configured hosting restoration otherwise look sound.

non-blocking: the duplicated read-only/lock/hosting state and the growing input policy in `video.vue` would benefit from consolidation, and an integration test for message → release → state change → acknowledgement would strengthen this boundary.

images/chromium-headful/client/src/components/video.vue
ehfeng: Fixed in 587dadf. Unlocking invalidates the modifier cache, and mouseleave no longer overwrites it while locked or while that invalidation is pending. The next mouseentry sends the current modifiers, then normal deduplication resumes. Regression tests cover Caps Lock, Num Lock, Scroll Lock, and unlocking while the pointer stays inside the viewer before leaving and reentering.

images/chromium-headful/client/tests/video-input.test.ts
ehfeng: Fixed in 587dadf. The live-view workflow now sets up Node 22 and runs npm ci using the client package-lock.json before bun test. Verified from a fresh dependency install: all 22 tests pass and the production client build succeeds.

masnwilliams (approved):
The latest commit addresses both requested changes: unlock now invalidates modifier state until the next valid synchronization point, with coverage for each lock key and the leave/reentry path; the live-view workflow now installs the pinned client dependencies before running Bun tests. I reran all 22 tests and the production build locally, and every current check including BugBot is green. No remaining blocking findings from me.

PR #393, opened 2026-09-16
Keep the OTLP export credential on the configured endpoint

archandatta:
## Invariant

**The authenticated export client attaches its credential only to requests for `BTEL_OTLP_ENDPOINT`.** A request to any other host sends no instance bearer token, and the authenticated exporter does not follow redirects away from the configured endpoint.

## Why

`bearerRoundTripper` resolves the bearer per request rather than capturing it once, which lets a forked VM export with the fresh instance JWT from its applied identity payload without restarting the process (#316). Because a `RoundTripper` runs on every hop after `net/http` applies its redirect-header policy, the credential needs its own explicit destination boundary.

Nothing discovers the export target. It is `BTEL_OTLP_ENDPOINT`, baked in at provisioning time, so a redirect is never something the authenticated exporter should act on and any host other than that endpoint is not the place the instance credential belongs.

## What changed

- `bearerRoundTripper` gains a `host`, set from `OTLPConfig.Endpoint`, and attaches the instance credential only when `req.URL.Host` matches it case-insensitively. A mismatch **fails closed**: no instance bearer token is added.
- The authenticated export client sets `CheckRedirect` to `http.ErrUseLastResponse`, so a `3xx` from the endpoint surfaces to the exporter as the response it is.

Explicitly unchanged: per-request credential resolution and the fork-refresh behavior it exists for, the request clone in `RoundTrip`, batch tuning, the `Headers` option, resource attributes, and the unauthenticated client path when `AuthTokenFunc` is unset. When `AuthTokenFunc` is set but returns an empty token, the request remains unauthenticated while the client still rejects redirects. A correctly configured authenticated exporter sends exactly the same request it sent before.

A `3xx` from the endpoint is surfaced by the OTel SDK as a non-retryable failure and counted in `kernel_otlp_export_failures_total`, where previously the client followed the `Location` response.

## Testing

Two new tests, both verified to fail with their respective guard reverted and to pass with it restored.

`TestBearerRoundTripper_AuthenticatesOnlyConfiguredHost` is table-driven over four cases: the configured host gets the credential, the same host uppercased still gets it, a different host gets nothing, and the same hostname on a different port gets nothing.

`TestOTLPStorageWriter_DoesNotFollowRedirects` stands up an endpoint that answers `302` toward a second server, waits for the terminal export failure, and asserts exactly one failed export, zero successful exports, one endpoint request, and zero redirect-target requests. Its cleanup stops the writer and releases the exporter.

- `go test ./lib/events/ -count=1`: 42 tests pass
- `go test ./lib/events/ -count=1 -race`: pass
- `go vet ./lib/events/`: pass
- `go build ./...`: pass

## Deployment

This lands in the browser image, so it reaches production on the next image release and version-pin bump rather than at merge.

---

> [!NOTE]
> **Medium Risk**
> Changes authenticated OTLP export behavior (credential attachment and redirect handling); misconfiguration could affect telemetry delivery, but the change fails closed on host mismatch.
> 
> **Overview**
> **Tightens OTLP export auth** so the instance JWT is only sent to the configured `BTEL_OTLP_ENDPOINT`, and redirects from that endpoint are not followed.
> 
> `bearerRoundTripper` now stores the configured endpoint host and adds `Authorization: Bearer …` only when the request host matches (case-insensitive); other hosts get no credential. The authenticated HTTP client sets `CheckRedirect` to `http.ErrUseLastResponse`, so a `3xx` from the endpoint is returned to the exporter as a failure instead of following `Location` (and potentially sending the token elsewhere).
> 
> Tests cover host/port matching for the round tripper and end-to-end behavior that a redirect is counted as one export failure with no request to the redirect target.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13c7b6b5b7fc343e240af9a7a50c963cb9f60175. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (approved):
reviewed — looks good.

The credential is now scoped to the configured OTLP host and port, authenticated exports reject redirects, and the tests cover both boundaries while preserving per-request token refresh behavior.

Validated with `go test ./lib/events/ -count=1`.

Sayan- (approved):

